PR #16, opened 2018-04-30
WIP: Documentation

AndreasAlbert:
Don't merge yet


clelange:
Looks great! What about moving the instructions before the HEPData definitions?
I didn't go through the full detail of each description, but will do so in the next couple of days. Anyway, this should definitely be merged, so go ahead.

PR #22, opened 2018-05-14
Fix pyyaml build

AndreasAlbert:
Works now, `pip3` was the key to success. Disabled `wrong-import-order` warning in `__init__.py`, which appeared on OSX, but not linux. I suspect it came up because yaml is now installed locally.

Regarding merge: Can just squash the commits in the merge.

### DIFF
--- a/.travis/setup.sh
+++ b/.travis/setup.sh
@@ -13,4 +13,5 @@ fi
 # Workaround for uggy PyYaml v 3.12 on pypi
 # Can be removed once PyYaml 3.13 is released
 mkdir ./yaml
+export PYTHONPATH="${PYTHONPATH}:$(readlink -e ./yaml)"
 easy_install --prefix="./yaml" git+https://github.com/yaml/pyyaml
--- a/.travis/setup.sh
+++ b/.travis/setup.sh
@@ -12,4 +12,4 @@ fi
 
 # Workaround for uggy PyYaml v 3.12 on pypi
 # Can be removed once PyYaml 3.13 is released
-pip install git+https://github.com/yaml/pyyaml
+easy_install git+https://github.com/yaml/pyyaml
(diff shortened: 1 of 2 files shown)


hepdata_lib/__init__.py
```diff
@@ -1,4 +1,5 @@
 """hepdata_lib main."""
+# pylint:disable=wrong-import-order
```
clelange: Is this really needed? What's the error from pylint?

.travis/setup.sh
```diff
@@ -5,7 +5,17 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     brew cask uninstall --force oclint
     brew upgrade python
     brew install root
+
+    # Workaround for buggy PyYaml v 3.12 on pypi
+    # Can be removed once PyYaml 3.13 is released
+    pip3 install git+https://github.com/yaml/pyyaml
```
clelange: In principle, it would be nicer to refer to a tag or hash instead of using the HEAD, but fine, it's (hopefully) just a temporary workaround

hepdata_lib/__init__.py
AndreasAlbert:
```
_______________________ [pylint] hepdata_lib/__init__.py _______________________
C: 11, 0: third party import "import yaml" should be placed before "import numpy as np" (wrong-import-order)
===================== 1 failed, 10 passed in 8.43 seconds ======================
```

.travis/setup.sh
AndreasAlbert: Agree, will improve.

hepdata_lib/__init__.py
clelange: Ah, OK, but then isn't it better to change the import order temporarily instead of risking to disable this check permanently?

AndreasAlbert: I did not explicitly try it, but I think the tricky bit is that this only happens in the osx build, but not on linux. So I figure that if I switch them around, it will stop working on linux. I can check if switching it works.

PR #25, opened 2018-05-22
Fix readthedocs build with autodoc

### LINKED ISSUE #23: readthedocs build failure: missing ROOT dependency.
For autodoc to work, we somehow need to make ROOT work on the readthedocs build machine.
The trouble is that AFAIK there is no way of executing a custom shell script before the build, so we cannot copy&paste the solution we use for travis.

One hacky solution would be to somehow cram the ROOT installation (e.g. via binary download as done with travis) into setup.py. In that case, we would have to make sure that this really only happens when installing the library on the build server and not for users.

AndreasAlbert:
Closes #23

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -160,3 +160,4 @@ texinfo_documents = [
 
 
 # -- Extension configuration -------------------------------------------------
+autodoc_mock_imports = ["ROOT"]
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,2 @@
 m2r
 sphinx_rtd_theme
-git+https://github.com/clelange/hepdata_lib
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 m2r
 sphinx_rtd_theme
+git+https://github.com/clelange/hepdata_lib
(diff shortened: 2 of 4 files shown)


AndreasAlbert:
It seemed to me that this is an internal travis problem rather than a problem with our code. I can install and run the package just fine on my personal machine in Python 2.7

AndreasAlbert:
Something is fishy here. 

When I install pylint in a clean virtual environment with python 2.7.12, it automatically installs these dependencies:
```
Successfully installed astroid-1.6.4 backports.functools-lru-cache-1.5 configparser-3.5.0 enum34-1.1.6 futures-3.2.0 isort-4.3.4 lazy-object-proxy-1.3.1 mccabe-0.6.1 pylint-1.9.1 singledispatch-3.4.0.3 six-1.11.0 wrapt-1.10.11
```

Notably including enum34, making me wonder why that package is missing in the travis build. 
After explicitly adding enum34 as a dependency to setup.py, the build fails with a SyntaxError that is due to a Python 2.7 / 3.x compatibility issue:

```
  File "/home/travis/build/clelange/hepdata_lib/.eggs/astroid-2.0.0.dev0-py2.7.egg/astroid/node_classes.py", line 649
    yield from ()
             ^
SyntaxError: invalid syntax
```

According to the pylint docs, compatibility should be fully handled....???

clelange:
I'll look into this using the travis CI python docker container

PR #26, opened 2018-05-23
Update documentation

clelange:
Making it work with readthedocs reducing the overall impact of mocking.

Docs are building: https://hepdata-lib.readthedocs.io/en/docu_mocking/ @AndreasAlbert does this look as expected?

Also trying to make tests work again, at least for linux it seems to work, for macosx the gcc installation seems to timeout a lot.

Closes #25 and #23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ matrix:
 
 
 before_install:
-  - ./.travis/setup.sh
+  - travis_retry ./.travis/setup.sh
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cd root; source bin/thisroot.sh; cd ..; fi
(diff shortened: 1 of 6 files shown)


clelange:
I found [this](http://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules) for the documentation and for the python packages I actually ran things locally using docker with the Travis CI image as documented [here](https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image). The retry feature I found on the [same page](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) (and then I searched the web to see if there's anything else that could be done, and some links pointed back to this).

AndreasAlbert:
Closes #23 #25

PR #27, opened 2018-05-24
Make RTD document the right version

AndreasAlbert:
In current master, RTD installs hepdata_lib from pypi and then shows documentation for that version, no matter which state of the repo it checks out. Fix by removing hepdata_lib from docs/requirements. There was no reason to have it in there in the first place, it's just something that stuck while I was testing stuff.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,2 @@
 m2r
 sphinx_rtd_theme
-hepdata_lib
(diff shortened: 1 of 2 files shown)


AndreasAlbert:
Merge at your convenience

AndreasAlbert:
Btw: It seems that even with `travis_retry`, a fraction of osx build attempts fails because `gcc` installation does not succeed immediately.

AndreasAlbert:
-> #28

Issue #20, opened 2018-05-08
Uncertainty setter does not allow second argument


AndreasAlbert:
With the current setup, there is no way to pass the `nominal` optional argument to the values setter function. See [here](https://github.com/clelange/hepdata_lib/blob/master/hepdata_lib/__init__.py#L379-L380)

AndreasAlbert:
Fixed in #29

PR #29, opened 2018-05-30
Improve 2D histogram reading and Uncertainty setter

AndreasAlbert:


AndreasAlbert:
While I was at it, I also fixed up the uncertainty setter function.

Closes #20

clelange:
Hi @AndreasAlbert - in principle, we should add the appropriate tests for this extension in [tests/test_uncertainty.py](https://github.com/clelange/hepdata_lib/blob/master/tests/test_uncertainty.py), shouldn't we? Otherwise, it's not guaranteed that it'll work also for python3.

AndreasAlbert:
good point

Issue #9, opened 2018-04-27
add TH1 reading functionality


clelange:
I have some example code available.

clelange:
Closed by #30

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #30, opened 2018-06-04
RootFileReader: Read TH1, improve TH2 reading + tests

AndreasAlbert:
* Added reader for TH1
* TH2 reader now also returns bin errors
* Tests for both

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -566,7 +566,6 @@ class RootFileReader(object):
     def read_hist_2d(self, path_to_hist):
         """Read in a TH2.
 
-
         :param path_to_hist: Absolute path in the current TFile.
         :type path_to_hist: str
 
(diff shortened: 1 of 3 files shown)


AndreasAlbert:
Closes #9

AndreasAlbert:
@clelange Can we merge this, please?

hepdata_lib/__init__.py
```diff
@@ -666,10 +679,42 @@ def get_hist_2d_points(hist):
...
+    points = {}
+    for key in ["x", "y", "x_edges", "dy"]:
+        points[key] = []
+
+    for x_bin in range(1, hist.GetNbinsX() + 1):
+        x_val = hist.GetXaxis().GetBinCenter(x_bin)
+        width_x = hist.GetXaxis().GetBinWidth(x_bin)
+
+        y_val = hist.GetBinContent(x_bin)
+        dy_val = hist.GetBinError(x_bin)
```
clelange: Shouldn't we use asymmetric errors here, i.e. `GetBinErrorLow()` and `GetBinErrorUp()`?

tests/test_rootfilereader.py
```diff
@@ -3,11 +3,13 @@
 """Test RootFileReader."""
 import random
 from unittest import TestCase
-import test_utilities
+from test_utilities import *
```
clelange: I'd prefer to avoid `import *`, is it really necessary?

hepdata_lib/__init__.py
AndreasAlbert: I didn't realize this existed. Added it now.

tests/test_rootfilereader.py
AndreasAlbert: You are right. Fixed now.

PR #38, opened 2018-06-15
Submission: New function for adding additional resource files.

AndreasAlbert:
Useful for the case where you want to add actual files to the tar ball, e.g. MC configuration cards.
As it is, this supersedes the functionality of the `add_link` function, but I have not removed it for now.

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -8,6 +8,7 @@ import math
 from collections import defaultdict
 import subprocess
 import warnings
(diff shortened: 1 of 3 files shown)


AndreasAlbert:
Added test for 
* nonexistant file -> RuntimeError
* existant, but too big -> RuntimeError
* existant and right size -> No RuntimeError

AndreasAlbert:
Ok, can be merged now @clelange

PR #56, opened 2018-07-26
Adding axis ranges support

rappoccio:
Hi Guys,

I have a feature request to add x and y axis ranges, if the user specifies them. It's needed to use this for SMP-16-010. 

Thanks,
Sal


AndreasAlbert:
@rappoccio Thanks for this contribution!

The code seems fine, but there are some style issues (white space, None comparisons, etc) that pylint picked out, which is why one of the tests fails. Please have a look at the [job log](https://travis-ci.org/clelange/hepdata_lib/jobs/408319183) (scroll all the way down) to see the line-by-line pylint output. Please implement these code style changes.

Optimally, we should also implement tests for this new functionality. Something like [the existing histogram tests](https://github.com/clelange/hepdata_lib/blob/master/tests/test_rootfilereader.py#L128)  but also using the axis range.

clelange:
I think there are a couple of new things that were recently added to pylint, which we should probably disable (http://pylint.pycqa.org/en/latest/whatsnew/2.0.html):

```
hepdata_lib/__init__.py:135:0: R0205: Class 'Variable' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance)
hepdata_lib/__init__.py:264:0: R0205: Class 'Table' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance)
hepdata_lib/__init__.py:413:0: R0205: Class 'Submission' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance)
hepdata_lib/__init__.py:582:0: R0205: Class 'Uncertainty' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance)
hepdata_lib/__init__.py:668:0: R0205: Class 'RootFileReader' inherits from object, can be safely removed from bases in python3 (useless-object-inheritance)
hepdata_lib/__init__.py:728:8: R1705: Unnecessary "else" after "return" (no-else-return)
hepdata_lib/__init__.py:769:4: R0913: Too many arguments (6/5) (too-many-arguments)
hepdata_lib/__init__.py:856:0: R0914: Too many local variables (20/15) (too-many-locals)
```

I'd suggest we disable `useless-object-inheritance` since it seems to be python3 only. 

@AndreasAlbert What do you think about `too-many-locals` and `too-many-arguments`? In principle, this would want us to use some other data structures instead, but I have no strong feeling about that - for now we could just adjust the thresholds (`max-args = 6` and `max-locals = 20` in https://github.com/clelange/hepdata_lib/blob/master/pylintrc#L505-#L517).

The `no-else-return` makes sense, so we'd have to remove it in https://github.com/rappoccio/hepdata_lib/blob/26f91643d53ea4161c84eecd835a72007a0b002d/hepdata_lib/__init__.py#L730

rappoccio:
Hi, Guys,

I have adjusted according to the style conventions, and added unit tests.

Cheers,
Sal

AndreasAlbert:
@clelange Too many locals does not make a lot of sense to me for short functions like the ones we are looking at here. I actually think that having all the bin indices in separate variables probably increased readability, so raising the threshold is probably the way to go. 

For arguments, I am not sure about the general case, but in this specific case, I would just go from `xmin=None, xmax=None` to using `xlim=(None,None)` as a single argument to save space, since it is probably more usable with fewer arguments.

IMO, we can also disable the `no-else-return`. Obviously pylint is correct that it is unnecessary and in this specific case, it's probably fine to skip the else. However, in general, I would prefer to not have to read and understand what happens in the `if` block to understand that there is an implied `else`. If there was more code, I would have to go looking through the `if` block to see if there's a return there. If there's a top-level `else`, it is immediately clear that the two blocks are exclusive.

clelange:
I've cleaned up the code in a few places and added a few more tests. The axis range parameters are now handled as `**kwargs` and also documented using the docstrings in sphinx (562685b).

The module has now reached > 1000 lines of code, so we should refactor into submodules. For now, I've set pylint to ignore `too-many-lines` (d9ed87f).

For the `get_hist_2d_points` function I think it's OK for now to just ignore that there are too many locals (8dfcda2).

tests/test_rootfilereader.py
```diff
@@ -158,6 +158,46 @@ def test_read_hist_1d_symmetric_errors(self):
...
+            hist.SetBinError(i+1, dy_values[i-xmin])
+
+        f = ROOT.TFile(fpath, "RECREATE")
+        hist.SetDirectory(f)
+        hist.Write("test")
+        f.Close()
+
+        reader = RootFileReader(fpath)
+
+        with self.assertRaises(TypeError):
```
AndreasAlbert: Add comment to clarify that this is a check for correct rejection of unknown kwargs

tests/test_rootfilereader.py
```diff
@@ -255,6 +295,73 @@ def test_read_hist_2d_symmetric_errors(self):
...
+        backup_hist = hist.Clone("backup")
+        rfile = ROOT.TFile(fpath, "RECREATE")
+        hist.SetDirectory(rfile)
+        hist.Write("test2d_sym")
+        rfile.Close()
+
+        reader = RootFileReader(fpath)
+
+        with self.assertRaises(TypeError):
+            reader.read_hist_1d("test", xlim=(xmin, xmax), ylim=(ymin, ymax), zlim=(ymin, ymax))
```
AndreasAlbert: Shouldn't this be calling `read_hist_2d` rather than `1d`?

tests/test_rootfilereader.py
```diff
@@ -255,6 +295,73 @@ def test_read_hist_2d_symmetric_errors(self):
...
+        with self.assertRaises(TypeError):
+            reader.read_hist_1d("test", zlim=(xmin, xmax))
```
AndreasAlbert: same

tests/test_rootfilereader.py
```diff
@@ -255,6 +295,73 @@ def test_read_hist_2d_symmetric_errors(self):
...
+
+        backup_hist = hist.Clone("backup")
+        rfile = ROOT.TFile(fpath, "RECREATE")
+        hist.SetDirectory(rfile)
+        hist.Write("test2d_sym")
+        rfile.Close()
+
+        reader = RootFileReader(fpath)
+
+        with self.assertRaises(TypeError):
```
AndreasAlbert: Please add comment as above

tests/test_rootfilereader.py
AndreasAlbert: What happens if I pass a known kwarg with nonsensical values? Like `xmin="Some string"`. We should test for that.

hepdata_lib/__init__.py
```diff
@@ -874,15 +874,15 @@ def get_hist_2d_points(hist,xmin=None,xmax=None,ymin=None,ymax=None):
     for key in ["x", "y", "x_edges", "y_edges", "z", "dz"]:
         points[key] = []
 
-    ixmin = hist.GetXaxis().FindBin(xmin) if xmin != None else 1
```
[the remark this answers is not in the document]
clelange: Github shows a weird diff now and I'm not sure I understand the comment. The current line reads:
```
ixmin = hist.GetXaxis().FindBin(xlim[0]) if xlim[0] is not None else 1
```
The idea is that the user passes `None` (or not even `xlim`) if no boundary should be used, and then the lower bin is the first bin. What am I missing?

.gitignore
```diff
@@ -106,4 +111,5 @@ ENV/
 .mypy_cache/
 
 # Virtual Studio Code
-.vscode/
\ No newline at end of file
+.vscode/
+.idea/
```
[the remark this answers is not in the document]
clelange: done

AndreasAlbert:
Thanks @clelange , looks good to me now.

Pylint is now only complaining about unnecessary parentheses. Once you fix these, the build should go through and we can merge this:
```
hepdata_lib/__init__.py:799:0: C0325: Unnecessary parens after 'assert' keyword (superfluous-parens)
hepdata_lib/__init__.py:802:0: C0325: Unnecessary parens after 'assert' keyword (superfluous-parens)
hepdata_lib/__init__.py:829:0: C0325: Unnecessary parens after 'assert' keyword (superfluous-parens)
hepdata_lib/__init__.py:932:0: C0325: Unnecessary parens after 'assert' keyword (superfluous-parens)
hepdata_lib/__init__.py:935:0: C0325: Unnecessary parens after 'assert' keyword (superfluous-parens)
hepdata_lib/__init__.py:1002:0: C0325: Unnecessary parens after 'assert' keyword (superfluous-parens)
```

clelange:
New version v0.2.3 built: https://pypi.org/project/hepdata_lib/0.2.3/

Issue #57, opened 2018-07-31
library needs refactoring


clelange:
The library has reached > 1000 lines and pylint is complaining, so we should move some of the classes into separated files and then enable `too-many-lines` (d9ed87f).

AndreasAlbert:
There are some obvious candidates:
* Refactor into something like `helpers.py`: `execute_command`, `find_all_matching`, `relative_round`, `check_file_existence`, `check_file_size`

* Refactor into something like `RootInteraction.py`: `RootFileReader`, `get_graph_points`, `get_hist_1d_points`, `get_hist_2d_points`

PR #58, opened 2018-08-02
Refactoring.

AndreasAlbert:
First stab at #57. Any thoughts @clelange?


AndreasAlbert:
Utility functions now live in `helpers` subspace. Everything to do with ROOT data formats moved to `RootUtils`.

clelange:
Hi Andreas,

In general, this looks very good to me. I think it's good to abstract away the classes that should not be used by the user to some extent.
Could you reenable the pylint option for the > 1000 lines file?

It's a little bit inconsistent to use CamelCase for `RooUtils` and then all lower case `helpers`, use snake_case for both? Then we would have a convention that classes that should be used by the user are CamelCase and the other snake_case?

AndreasAlbert:
Oh yeah, let's just use snake for both.

Guiding the user with the case convention used in the code is a bit too arcane for my taste.

AndreasAlbert:
To be more precise:
CamelCase for class names, snake for everything else (which is what we did so far, I think)

clelange:
I think there's only one error left:
```
hepdata_lib/__init__.py:258:12: E0602: Undefined variable 'execute_command' (undefined-variable)
```

AndreasAlbert:
* Fixed pylint errors
* Applied snake_case to RootUtils submodule -> root_utils
* Travis now runs pylint on all python sub files

hepdata_lib/__init__.py
```diff
@@ -22,6 +22,8 @@
 import numpy as np
```
clelange: We actually don't need the ROOT and numpy imports here anymore, do we?

AndreasAlbert: Yes, removed.

AndreasAlbert:
You are right about the backward compatibility. For now, I added an import to the main file as per your suggestion. Longer term, we should encourage users to switch their import statements.

clelange:
I'll work on #60 first so that we have a working docker container and binder examples again. Will get to it this week.

AndreasAlbert:
If #60 is not a code issue on our side, why does this PR need to wait for it?

PR #59, opened 2018-08-06
Add more testing for RootFileReader class

AndreasAlbert:
Only merge after #58

### DIFF
--- a/tests/test_rootfilereader.py
+++ b/tests/test_rootfilereader.py
@@ -25,15 +25,13 @@ class TestRootFileReader(TestCase):
 
         # Check with existant file that does not end in .root
         path_to_file = "test.txt"
+        self.addCleanup(os.remove, path_to_file)
+
         with open(path_to_file,"w") as testfile:
             testfile.write("TEST CONTENT")
 
         with self.assertRaises(RuntimeError):
             reader = RootFileReader(path_to_file)
-
-        # Clean up
-        os.remove(path_to_file)
-
 
         # Check with wrong input type
         with self.assertRaises(ValueError):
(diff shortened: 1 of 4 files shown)


clelange:
@AndreasAlbert 
Conflicting file: tests/test_rootfilereader.py

AndreasAlbert:
Hi @clelange the merge conflicts have been resolved. It seems CI seems to fail because the OSX build times out, as it does in other PRs. Can we please merge this?

tests/test_rootfilereader.py
```diff
@@ -7,13 +7,44 @@
...
+        # Check with wrong input type
+        with self.assertRaises(ValueError):
+            reader = RootFileReader(5)
+        with self.assertRaises(ValueError):
+            reader = RootFileReader([])
+        with self.assertRaises(ValueError):
+            reader = RootFileReader({})
+
+        # Finally, try a good call
+        testfile = ROOT.TFile("testfile.root")
```
clelange: Can you use `path_to_file = make_tmp_root_file(close=True)` and delete the ROOT file at the end again? And also delete `test.txt`?

AndreasAlbert:
I overhauled the output cleanup completely. It is now done using the `TestCase.addCleanup` method. This allows you to register a clean-up action that will be taken automatically, no matter if there is an exception or the test fails. At the end of each test method, `TestCase.doCleanups` is called to ensure that the files are not only cleaned up at the end of all tests, but also in between tests. Finally, I changed all temporary root files to use the dedicated function and made it so that the function automatically registers the temporary files for deletion.

PR #62, opened 2018-08-27
Fixed typo affecting calculation of y_edges.

aehart:
This is a pretty simple change that is necessary to handle axes with nonuniform bin edges.

### DIFF
--- a/hepdata_lib/root_utils.py
+++ b/hepdata_lib/root_utils.py
@@ -282,7 +282,8 @@ def get_hist_2d_points(hist, **kwargs):
     symmetric = (hist.GetBinErrorOption() == r.TH1.kNormal)
     for x_bin in range(ixmin, ixmax):
         x_val = hist.GetXaxis().GetBinCenter(x_bin)
-        width_x = hist.GetXaxis().GetBinWidth(x_bin)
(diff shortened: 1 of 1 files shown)


AndreasAlbert:
Hi @aehart, thank you for contributing! Could you clarify for what case this change is necessary? What do you mean by "nonuniform bin edges"? At first I thought you were referring to non-uniform bin sizes, but the old method seems to work fine for those, see [this quick test](https://gist.github.com/AndreasAlbert/0ad39279649950f8b89b23b462ad2420).

It would be a good idea to add a test case for your scenario to make sure we get it right in the future (either in the existing [RootFileReader tests](https://github.com/clelange/hepdata_lib/blob/master/tests/test_rootfilereader.py) or in a separate test group for the `root_utils` submodule).

aehart:
Hi @AndreasAlbert. The issue I was encountering turned out to be due to a simple typo, as opposed to a problem with how the bin edges were calculated. The x-axis was being used to get width_y, which was leading to nonsensical y_edges for me.

Sorry about that. I just ploughed through the code with a solution to what I initially thought was the problem without thinking too carefully. I've updated the PR to use the original method for calculating bin edges, but with the typo fixed.

AndreasAlbert:
Oh yes, clearly a typo. Good catch!

PR #64, opened 2018-08-29
Fix MacOS ROOT python3 installation for TravisCI

clelange:
Basically I'm now just installing all ROOT dependencies first, then download a tarball of my installation from CERN and move that into the right place. Installing via homebrew takes more than the allowed 50 minutes since ROOT needs to be compiled from source...

The `travis_wait` stuff is nice to have (see https://github.com/clelange/hepdata_lib/compare/travis_wait?expand=1#diff-a677667c17d131d7fe128bf3b226afe9R14) for usage, not needed for this PR, but could be useful at a later stage for long-running tests.

Please squash-merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cd root; source bin/thisroot.sh; cd ..; fi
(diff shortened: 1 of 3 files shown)


AndreasAlbert:
Thanks for preparing this, @clelange! Looks good to me.

PR #70, opened 2018-11-29
add option to force symmetric errors to be read in from histograms

clelange:
Feature requested by @rappoccio - does not hurt and makes things more flexible.


AndreasAlbert:
Would it make sense to rather have a three-way option like
```
error_type = "auto" (default)
error_type = "sym"
error_type = "asym"
```
so that the use can also force asymmetric up/down errors?

clelange:
Hmm, OK, but forcing asymmetric up/down errors would only apply for cases where the errors in the histogram are symmetric, and then one would have to "patch" them manually after reading them in to become asymmetric.

We only use symmetric errors when the histograms itself has `hist.GetBinErrorOption() == r.TH1.kNormal`, which quoting from the [ROOT TH1 documentation](https://root.cern.ch/doc/master/classTH1.html#ac6e38c12259ab72c0d574614ee5a61c7) means:

enumerator | description
------------| ------------
kNormal | errors with Normal (Wald) approximation: errorUp=errorLow= sqrt(N)
kPoisson | errors from Poisson interval at 68.3% (1 sigma)
kPoisson2 | errors from Poisson interval at 95% CL (~ 2 sigma)

If you think there is a realistic use case I can implement this. Let me know.

PR #72, opened 2018-12-13
If no unit is given, no '[]' is printed.

reimersa:


hepdata_lib/__init__.py
```diff
@@ -134,7 +134,10 @@ def make_dict(self):
         Except for debugging purposes, no user should have to call this function.
         """
         tmp = {}
-        tmp["header"] = {"name": self.name, "units": self.units}
+        if self.units != '':
```
AndreasAlbert:
```suggestion
        tmp["header"] = {"name": self.name}
        if self.units:
            tmp["header"]["units"] = self.units
```

hepdata_lib/__init__.py
```diff
@@ -134,7 +134,10 @@ def make_dict(self):
...
+            tmp["header"] = {"name": self.name, "units": self.units}
```
AndreasAlbert:
```suggestion
            
```

hepdata_lib/__init__.py
```diff
@@ -134,7 +134,10 @@ def make_dict(self):
...
+        else:
```
AndreasAlbert:
```suggestion

```

hepdata_lib/__init__.py
```diff
@@ -134,7 +134,10 @@ def make_dict(self):
...
+            tmp["header"] = {"name": self.name}
```
AndreasAlbert:
```suggestion

```

AndreasAlbert:
Hi @reimersa, thanks for this contribution! The empty "[]" thing is indeed quite annoying. I would propose to slightly change your suggested code to avoid the duplication of the assignment line. I am not sure whether this suggestion will show up correctly with the in-code github comment, so here it is again in plain text:

```
        tmp["header"] = {"name": self.name}
        if self.units:
            tmp["header"]["units"] = self.units
```

Let me know what you think.

reimersa:
Hi @AndreasAlbert ,

sure, that makes perfect sense to me, thanks for the suggestion!

PR #73, opened 2018-12-13
More testing

AndreasAlbert:


AndreasAlbert:
I added testing for the `add_image` and associated function, as well as a few type or input checks here and there. The apparent ImageMagick problem on travis was solved by installing ghostscript, which was simply missing on the workers.

clelange:
Thanks a lot, getting our test coverage up by almost 10% is great!

One comment: I would suggest to add a small PDF file (I found one that's < 1 kB) to the repository instead of trying to search for one locally. If you agree, I can take care of it.

AndreasAlbert:
Sure, go ahead.

PR #74, opened 2018-12-14
remove trailing whitespace, add __version__ property

clelange:
Make pylint happy again and tests pass.


clelange:
Also added `__version__` property so that people can just print out the version they are using in case they cannot run `pip`

PR #77, opened 2018-12-17
Do not restrict uncertainty signs

intrepid42:
Uncertainty signs should not be restricted as a variation can yield different signs in different parts of he distribution, especially (but not only) when normalized.

I did not check the tests, let's see if it works!

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -531,24 +531,8 @@ class Uncertainty(object):
 
         """
         if self.is_symmetric:
-            try:
-                assert all([x >= 0 for x in values])
-            except AssertionError:
-                raise ValueError(
-                    "Uncertainty::set_values: Wrong signs detected!\
-                     For symmetric uncertainty, all uncertainty values should be >=0."
-                )
             self._values = values
         else:
-            try:
-                assert all([x[1] >= 0 for x in values])
-                assert all([x[0] <= 0 for x in values])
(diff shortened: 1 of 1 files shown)


AndreasAlbert:
Hi @intrepid42, thanks for contributing!

I see your point for asymmetric uncertainties, but I do not see how it applies to symmetric uncertainties. In case of symmetric uncertainties, what would you mean by "uncertainy < 0" ?

intrepid42:
Hi Andreas, I think symmetric uncertainties are quoted to look nicer, or to obtain a more stable result when the statistical precision is limited. For these symmetrized uncertainties, the same argument as above applies: we do not want the uncertainty to "reflect" at 0 because this would change a variation of the mean into a variation of the resolution (for a "peak"-like observable).

image

AndreasAlbert:
Oh yeah, I'm an idiot. Makes sense. Thanks for the graphic!

intrepid42:
No problem. Sometimes a picture is worth a thousand words :)

intrepid42:
Thanks for merging!

Just to add: I updated the entry for TOP-17-013 to make sure that HepData accepts negative (positive) values for plus (minus) errors.

So we are good to go for building covariance matrices from the HepData error-split, see https://indico.cern.ch/event/779259/contributions/3242594/attachments/1770317/2876299/LCorpe_LHCEWWG_Correlations_131218.pdf

PR #82, opened 2019-02-11
Documentation

AndreasAlbert:
I added some documentation on the thumb nail feature and expanded the coverage of autodoc. Since they are pretty similar topics, the documentation on thumb nails and keywords now lives togeher under the `Tables and Variables` topic.


docs/usage.rst
```diff
@@ -152,6 +152,38 @@ That's it! We have successfully created the Table and Variables and stored our r
...
+::
+
+    table.add_image("path/to/image.pdf")
+
+The library code then takes care of all the necessary steps, like converting the image to the right format and size, and copying it into your submission folder. The conversion relies on the ImageMagick library, and will only work if the ``convert`` command is available on your machine.
+
+Adding keywords to a table
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To make hepdata entries more searchable, keywords can be used to define what information is shown in a table.
```
clelange: What about writing "should" instead of "can" since without the keywords, the entry will basically only show up when browsing or searching for the correct title? Or alternatively write something like "Please make sure that you/It is good practice to add a reasonable number of keywords to make your analysis show up when using the HEPData search and filters."

AndreasAlbert: Good proposal. Implementation see below.

PR #83, opened 2019-03-19
Fix pypi badge

AndreasAlbert:
The link needs a hyphen "-" rather than an underscore "_" to work properly.


clelange:
I've restarted the CI to see if we need to fix anything. I had actually contacted the fury people about this, but I never realised that this changed from underscore to hyphen...

PR #85, opened 2019-04-02
Improve variable constructor

AndreasAlbert:
The `values` property can now be set from the constructor, like:

``var = Variable("name", is_binned=False, values=[1,2,3])``

Additionally, the `values` setter now checks that the input format agrees with the `is_binned` property. If a Variable is binned, the input should be a list of two-valued tuples. If it is not, it should be parse-able into a list of single values.

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -53,7 +53,8 @@ class Variable(object):
     # pylint: disable=too-many-instance-attributes
     # Eight is reasonable in this case.
 
-    def __init__(self, name, is_independent=True, is_binned=True, units="", values=[]):
+    def __init__(self, name, is_independent=True, is_binned=True, units="", values=None):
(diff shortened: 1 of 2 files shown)


hepdata_lib/__init__.py
```diff
@@ -74,9 +74,21 @@ def values(self):
     def values(self, value_list):
         """Value Setter."""
         if self.is_binned:
+            # Check that the input is well-formed
+            try:
+                assert(all([len(x)==2 for x in value_list ]))
```
JoRoemer: As far as I know, assert will raise an AssertionError which will not be caught by the except block. So I would either catch the AssertionError, use the assert <assertion>, <error msg> syntax or simply use an if statement to raise the error.

AndreasAlbert: Yes, that's right. Should catch that, too.

PR #90, opened 2019-04-13
Limit the length of table names to 64 characters.

### LINKED ISSUE #34: Limit table name length to 64 characters
This limit is enforced by the hepdata servers. Trying to upload a submission with a table with name length > 64 characters will fail. We should prevent users from doing that.

AndreasAlbert:
Closes #34

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -203,7 +203,10 @@ class Table(object):
     the location within the paper, etc.
     """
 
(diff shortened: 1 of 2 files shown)


clelange:
Sorry about the delay, I'll get to this within the next couple of days.

clelange:
We could think about changing the pylint default in our config to allow more parameters in general:

```
max-attributes=10
```

PR #91, opened 2019-05-10
relative_round function does not handle tuples

drberry85:
Updated version of relative_round to handle tuples.

### DIFF
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -49,20 +49,30 @@ def find_all_matching(path, pattern):
     return result
 
 
-def relative_round(value, relative_digits):
-    """Rounds to a given relative precision"""
-    if value == 0:
-        return 0
-    if isinstance(value, str) or np.isnan(value):
-        return value
-
-    value_precision = math.ceil(math.log10(abs(value)))
-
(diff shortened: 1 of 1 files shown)


clelange:
Hi @drberry85 - thanks for catching this an the PR!
Do you think you could add a few tests corresponding to this issue as well?

drberry85:
Sure, where can I find the script that Travis runs. I just need add test data that has uncertainty as a tuple.

AndreasAlbert:
Hi @drberry85, travis is configured with the [travis.yaml file](https://github.com/HEPData/hepdata_lib/blob/master/.travis.yml#L19-L20), and you can see it simply runs tests by calling `python setup.py test`. This will execute all tests defined in the [test folder](https://github.com/HEPData/hepdata_lib/tree/master/tests), so in order to create new tests, you can add a test case to one of the existing `test_*.py` files, or in this case create a new file `test_helpers.py` since we do not have any tests for the helper functions, yet. You need to define a test class that inherits from `TestCase`, you can check out the existing tests for inspiration, e.g. [here](https://github.com/HEPData/hepdata_lib/blob/master/tests/test_submission.py).

AndreasAlbert:
We should document how to add tests, see #93

PR #96, opened 2019-05-24
Add testing documentation.

### LINKED ISSUE #93: Document how to add a new test
We should have clear documentation so that contributors know how to to it.

AndreasAlbert:
Closes #93.

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -1,0 +1,104 @@
+Developer information
+======================
+
+The testing system
+--------------------
+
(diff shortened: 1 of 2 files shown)


clelange:
Thanks a lot for writing this important piece of documentation! It looks good to me, but I was wondering whether we should explicitly mention `pytest` and not only just `unittest`.

Also, what do you think about linting the test code as well? I can work on it this week, but if we want more people to work on it, we should probably also enforce some coding standards.

PR #104, opened 2019-08-22
Add .readthedocs.yml

clelange:
This is an attempt to fix the build errors for the docs.
Documentation for this file: https://docs.readthedocs.io/en/stable/config-file/v2.html

I actually realised that also just changing the CPython version from 2.X to 3.X in the settings fixes the issue, but with this config we should be on the safe side.

Build successful: https://readthedocs.org/projects/hepdata-lib/builds/


AndreasAlbert:
Looking good, thanks!

PR #134, opened 2020-04-21
Fix macOS tests

clelange:
The [homebrew ROOT formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/root.rb) uses `python@3.8` explicitly, so we need to do the same. Installing ROOT from source takes as in the past 45 minutes, so I chose to download a tar ball again.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,18 +26,14 @@ jobs:
       - uses: actions/checkout@v2
       - name: Test with pytest
         run: |
+          ROOT_VERSION=6.20.04_1
           brew update
-          brew install root
-          # brew cask uninstall --force oclint
-          # brew upgrade python cmake libpng libtiff
-          # brew install --force ossp-uuid davix fftw fontconfig gd graphviz gsl lz4 tbb xrootd
-          # curl -O https://clange.web.cern.ch/clange/root-v6.18.00.tar.gz
-          # tar xzf root-v6.18.00.tar.gz
-          # mkdir -p /usr/local/Cellar/root
-          # mv 6.18.00 /usr/local/Cellar/root/
-          ROOT_CELLAR=$(brew --cellar root)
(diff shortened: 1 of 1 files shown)


clelange:
There were some issues with GitHub in general (server issues), but now all the tests pass again.

AndreasAlbert:
I'm not very familiar with macos, so let me summarize what I think you're doing here:
1. you installed root using homebrew on a mac machine
2. you packed up the resulting binaries
3. you ship the binaries to the worker

And because step (1) is tied to python 3.8, you also need to use 3.8 for step (3)? Why is it necessary to use homebrew in the first place? Instead, could you replace step (1) by compiling from source using whatever python version you like? I don't have the contextual knowledge to judge whether that would be more or less straightforward.

clelange:
Yes, one could compile ROOT from source on macOS. This takes something like 45 minutes, and would have to be done for each test. Homebrew provides ROOT and I think that most people on macOS install it in this way. In most cases one just downloads the binaries and is done. However, due to the setup of the machine provided by GitHub, even when using Homebrew, ROOT is compiled from scratch again (i.e. 45 minutes of waiting). I don't know why. Independently, it requires Python 3.8. The only way I found to work around this is to simply tar up my ROOT installation and download it when running the tests.

clelange:
@AndreasAlbert I don't have a better suggestion than this. Shall we just move ahead with it so that at least for now the tests work? I would expect that this approach will continue to work for quite some time.

AndreasAlbert:
@clelange Apologies, I thought I had already merged this, your explanation made sense to me.

clelange:
Thanks! Now there is a new failure in the linux python3 tests, but I think this can be fixed by appending the `PYTHON_PATH` such that it "sees" the local `hepdata_lib` directory. I'll look into it in the next few days.

AndreasAlbert:
I think #136 might fix this

PR #135, opened 2020-04-29
Added possibility to read TGraph from '.C' file

SirVheod:
Added c_file_reader.py to be able to read normal TGraphs from .C file. Also added 'test_cfilereader.py' tests and a 'read_c_files.ipynb' notebook. 

Issue: add possibility to read TGraph from `.C` file #8

### DIFF
--- a/hepdata_lib/c_file_reader.py
+++ b/hepdata_lib/c_file_reader.py
@@ -1,15 +1,10 @@
 """.C file reader"""
 
-import collections
-from collections import defaultdict
-import numpy as np
-import hepdata_lib
-import ROOT
 import io
-from ROOT import TGraph
 from array import array
-from hepdata_lib.helpers import check_file_existence
+from ROOT import TGraph
 import hepdata_lib.root_utils as ru
+from hepdata_lib.helpers import check_file_existence
 
 class CFileReader(object):
     """Reads TGraphs from '.C' files"""
@@ -53,44 +48,43 @@ class CFileReader(object):
     def get_graphs(self):
         """Function to read the .C file"""
 
-        found_graphs = self.find_Graphs()
+        found_graphs = self.find_graphs()
         graphs = found_graphs[0]
         graph_names = found_graphs[1]
-        y_values = [ 'd' ]
-        x_values = [ 'd' ]
+        y_values = ['d']
+        x_values = ['d']
         list_of_tgraphs = []
         count = 0
 
         while count < len(graphs) -1:
-            name = graphs[count]
-            x_values = self.read_Graph(graphs[count])  
-            y_values = self.read_Graph(graphs[count+1])
-            tgraph = self.create_TGraph(x_values, y_values)
+            x_values = self.read_graph(graphs[count])
+            y_values = self.read_graph(graphs[count+1])
+            tgraph = self.create_tgraph(x_values, y_values)
             tgraph = dict(tgraph)
             list_of_tgraphs.append(tgraph)
             count += 2
 
         graph_object = zip(graph_names, list_of_tgraphs)
-        all_Graphs = dict(graph_object)
+        all_graphs = dict(graph_object)
 
-        return all_Graphs
+        return all_graphs
 
-    def create_TGraph(self, x, y):
+    def create_tgraph(self, x_value, y_value):
         """Function to create pyroot TGraph object"""
+        # pylint: disable=no-self-use
+        x_values = array('d')
+        y_values = array('d')
+        length = len(x_value)
 
-        x_values = array( 'd' )
-        y_values = array( 'd' )
-        length = len(x)
-
-        for value in range( length ):
-            x_values.append(x[value])
-            y_values.append(y[value])
-        t_object = TGraph( length, x_values, y_values )
+        for value in range(length):
+            x_values.append(x_value[value])
+            y_values.append(y_value[value])
+        t_object = TGraph(length, x_values, y_values)
         graph = ru.get_graph_points(t_object)
 
         return graph
 
-    def find_Graphs(self):
+    def find_graphs(self):
         """Find all TGraphs in .C file"""
 
         c_file = self.cfile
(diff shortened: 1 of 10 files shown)


hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
+""".C file reader"""
+
+import io
+from array import array
+from ROOT import TGraph
+import hepdata_lib.root_utils as ru
+from hepdata_lib.helpers import check_file_existence
+
+class CFileReader(object):
+    """Reads TGraphs from '.C' files"""
```
clelange, line 10: --> Reads ROOT Objects from .C files

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+        """The .C file this reader reads from."""
+        return self._cfile
+
+    @cfile.setter
+    def cfile(self, cfile):
+        """
+        Define the '.C' file to read from.
+        """
+        if isinstance(cfile, str):
+            if not cfile.endswith(".C"):
```
clelange, line 32: This should be case insensitive

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+                raise RuntimeError(
+                    "CFileReader: Input file is not a .C file (name does not end in .C)!"
+                    )
+            if check_file_existence(cfile):
+                self._cfile = open(cfile, "r")
+        elif isinstance(cfile, io.TextIOBase):
+            self._cfile = cfile
+        else:
+            raise ValueError(
+                "CFileReader: Encountered unkonown type of variable passed as cfile argument: "
```
clelange, line 42: Typo "unknown"

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+                + str(type(cfile)))
+
+        if not self._cfile:
+            raise IOError("CFileReader: File not opened properly.")
+
+    def get_graphs(self):
+        """Function to read the .C file"""
```
clelange, line 49: --> Parse the .C file trying to find TGraph objects

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+
+        found_graphs = self.find_graphs()
+        graphs = found_graphs[0]
+        graph_names = found_graphs[1]
+        y_values = ['d']
```
clelange, line 54: This makes the assumption all `TGraph` entries are of type float/double, but `TGraph` also has a constructor with only integers as values. Parsing them as float/double would lead to weird numbers. What about checking the values  after they have been read to see if they are entirely of type integer and then use this accordingly?

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+
+        c_file = self.cfile
+        values = []
+        start = 0
+        ignore = 0
+        c_file.seek(0, 0)
+
+        # Finding values from the correct object
+        for line in c_file.readlines():
+            if line.startswith('/*') or '/*' in line:
```
clelange, line 140: I think you should not ignore the full line, but only after the `/*`

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+                ignore = 1
+                continue
+            if '*/' in line:
```
clelange, line 143: Similar as above: start to not ignore after the `*/`

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+                ignore = 0
+                continue
+            if line.startswith('//'):
+                continue
+            if((graphname in line) and ("{" in line) and
+               (ignore == 0) and (line.startswith('//') is False)):
```
clelange, line 149: You don't need to check for `//` here and further below, it's done above already.

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+                start = 1
+                continue
+            if "}" in line and line.startswith('//') is False:
+                if start == 1:
+                    values.append(float(line.split("}")[0]))
```
clelange, line 154: If there's something wrong with the TGraph, the code will throw an exception, but it's not clear why. Could you think about a way that would first try to check whether this is actually a number that can be converted, and if this isn't the case, throw an error with the current line number in the .C file and the TGraph you're looking at?

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+
+        c_file = self.cfile
+        names = []
+        objects = []
+        graphs = []
+        start = 0
+        ignore = 0
+
+        for line in c_file.readlines():
+            if line.startswith('/*') or '/*' in line:
```
clelange, line 99: I looked first at `read_graph`, and the same comments apply. However, since the logic is the same, it would be better to actually define a function to which you pass the current line, which then returns the corresponding `ignore` value and e.g. a boolean that states whether to continue or not.

hepdata_lib/c_file_reader.py
```diff
@@ -0,0 +1,159 @@
...
+                ignore = 0
+                continue
+            if line.startswith('//'):
+                continue
+            if(("TGraph(" in line) and ("(" in line) and
+               (ignore == 0) and (line.startswith('//') is False)):
+                start = 1
+                objects.append(line.split('(', 1)[1].split(')')[0])
+                continue
+            if(("SetName(" in line) and ("(" in line) and
```
clelange, line 112: Here you check for "SetName", but what if the `TGraph` is simply not named? In this case, you would return `graphs` and `names` with different lengths? If there is no name set, I would suggest to assign the variable name of the `TGraph`.

hepdata_lib/c_file_reader.py
SirVheod: This part changed. It throws an error with the connected TGraph but not the line.

PR #141, opened 2020-05-15
Fixed issue with c_file_reader.py and added tests

SirVheod:
Fixed c_file_reader.py for not being able to read certain valid TGraphErrors.
- Removed unnecessary if statements from create TGraph and TGraphErrors functions
- Added few more test cases.

### DIFF
--- a/hepdata_lib/c_file_reader.py
+++ b/hepdata_lib/c_file_reader.py
@@ -93,15 +93,10 @@ class CFileReader(object):
         while count < len(graphs) -1:
             xvalues = self.read_graph(graphs[count])
             yvalues = self.read_graph(graphs[count+1])
-
-            try:
-                for value in xvalues:
-                    x_values.append(value)
-                for value in yvalues:
-                    y_values.append(value)
-            except ValueError:
-                raise IndexError("Invalid values. Int or float required.")
-
+            for value in xvalues:
+                x_values.append(value)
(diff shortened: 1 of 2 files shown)


hepdata_lib/c_file_reader.py
```diff
@@ -93,21 +93,15 @@ def create_tgraph_dict(self, graph_list, list_of_tgraphs):
...
-                     or any(not isinstance(x, int) for x in yvalues)):
-                    for value in xvalues:
-                        x_values.append(float(value))
-                    for value in yvalues:
-                        y_values.append(float(value))
+                for value in xvalues:
+                    x_values.append(value)
+                for value in yvalues:
+                    y_values.append(value)
             except ValueError:
```
clelange, line 102: Why don't you check for int/float anymore? I think that there is no possibility of getting a `ValueError` with the updated code, is there? You simply append to a list, so you can remove the `try` construct here. You check for this in https://github.com/HEPData/hepdata_lib/blob/79a324f106aa47da610762a2c31bec265c259d45/hepdata_lib/c_file_reader.py#L363-L404 - right?

SirVheod: Checks for float and int are made in the create_tgrapherrors and create_tgraph functions, so everything seems to work as they should even when these are removed. I removed the useless try checks.

PR #142, opened 2020-05-15
Improved tests.yml and added notebook previews Closes #124

### LINKED ISSUE #124: Preview notebooks once tested
Once the notebooks are tested (see #65), we can also write them out as artifacts to preview them:

```shell
jupyter nbconvert [--execute] --to html notebook.ipynb
```

SirVheod:
- Divided tests in tests.yml into steps.
- Added steps for converting executed notebooks into .html and saving them as artifacts. Notebooks can be downloaded after the workflow run has completed for preview.

Closes: #124 (Preview notebooks once tested)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,13 +26,13 @@ jobs:
             
         # Save notebooks
         - name: Save notebooks
-          if: always()
+          if: ${{ always() }}
           run: |
             python -m jupyter nbconvert --ExecutePreprocessor.timeout=600 --ExecutePreprocessor.allow_errors=True --to html --execute examples/*.ipynb
             
         # Upload notebooks
         - name: Upload notebooks
-          if: always()
+          if: ${{ always() }}
           uses: actions/upload-artifact@v2
           with:
             name: notebooks py3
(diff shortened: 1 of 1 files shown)


clelange:
Thanks for this! When I look at the newly generated artifacts, I see that the JupyROOT version is different and that makes me believe that the notebooks are not properly executed, but only converted? Could you please check?

SirVheod:
> Thanks for this! When I look at the newly generated artifacts, I see that the JupyROOT version is different and that makes me believe that the notebooks are not properly executed, but only converted? Could you please check?

yes they were missing --execute parameter as I thought they would use the already executed notebook but no

SirVheod:
> Thanks for this! When I look at the newly generated artifacts, I see that the JupyROOT version is different and that makes me believe that the notebooks are not properly executed, but only converted? Could you please check?

Now I got the tests work again and added "allow_errors" so that it actually saves the notebooks that fail to execute properly.

clelange:
Have you tested that even though you have `if: always()` set that the status will still be shown as failed in case there is something wrong with the tests?

Also, I think the documentation suggests to use braces, i.e. `if: {{ always() }}`, see https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always - it's probably better to follow that example.

PR #144, opened 2020-05-20
adding functions to manage rounding

cippy:
All changes in **hepdata_lib/helpers.py**
These functions are mostly utilities to facilitate rounding of values stored in dictionaries as returned, e.g., by RootFileReader::read_hist_1d()
The main one is roundValueAndUncertainty(), which allows to round values according to the significant digits of their respective uncertainties (rounded to a given input precision).

### DIFF
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -110,7 +110,7 @@ def get_value_precision_wrt_reference(value, reference):
             raise ValueError("Unsupported input type passed to " + this_function)
 
     return get_number_precision(value) - get_number_precision(reference)
-    
+
 
 def round_value_to_decimals(cont, key="y", decimals=3):
     """
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -133,7 +133,7 @@ def round_value_and_uncertainty_to_decimals(cont, val_key="y", unc_key="dy", dec
         cont[val_key][i] = round(val, decimals)
         if isinstance(unc, tuple):
             cont[unc_key][i] = (round(unc[0], decimals), round(unc[1], decimals))          
-       else:
+        else:
             cont[unc_key][i] = round(unc, decimals)
 
 
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -75,7 +75,7 @@ def get_number_precision(value):
     return math.ceil(math.log10(abs(value)))
 
 
-def get_value_precision_wrt_reference(value,reference):
+def get_value_precision_wrt_reference(value, reference):
     """
     relative precision of first argument with respect to the second one 
     value and reference must be of the same type
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -50,46 +50,67 @@ def find_all_matching(path, pattern):
     return result
 
 
-def relative_round(value, relative_digits):
-    """Rounds to a given relative precision"""
+def get_number_precision(value):
+    """
+    Get precision of an input value.
+    Exact integer powers of 10 are assigned same precision of smaller numbers
+    For example
+    get_number_precision(10.0) = 1
+    get_number_precision(10.001) = 2
+    get_number_precision(9.999) = 1
+    """
 
+    if isinstance(value, tuple):
+        return (get_number_precision(x) for x in value)
+
+    # if value is tuple, value == 0 might cause ValueError saying that
+    # 'The truth value of an array with more than one element is ambiguous'
+    # Thus, tuple is evaluated above
     if value == 0 or isinstance(value, str) or np.isnan(value) or np.isinf(value):
         return value
+
+    return math.ceil(math.log10(abs(value)))
+
+
+def relative_round(value, relative_digits):
+    """Rounds to a given relative precision"""
 
     if isinstance(value, tuple):
         return (relative_round(x, relative_digits) for x in value)
 
-    value_precision = math.ceil(math.log10(abs(value)))
+    if value == 0 or isinstance(value, str) or np.isnan(value) or np.isinf(value):
+        return value
+
+    value_precision = get_number_precision(value)
     absolute_digits = -value_precision + relative_digits
 
     return round(value, int(absolute_digits))
 
 
-def get_number_precision(value):
-    """Get precision of an input value."""
-
-    if value == 0 or isinstance(value, str) or np.isnan(value) or np.isinf(value):
-        return value
-    if isinstance(value, tuple):
-        return (get_number_precision(x) for x in value)
-
-    return math.ceil(math.log10(abs(value)))
-
-
 def get_value_precision_wrt_reference(value, reference):
     """
     relative precision of first argument with respect to the second one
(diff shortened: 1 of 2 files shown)


clelange:
Hi @cippy, thanks a lot for this contribution! The tests show some failures, one related to the rounding, the other one simply being a linter error. Do you think you will be able to fix them yourself or would you like some help with it?

Do you think you could come up with some tests for the new functions? Also here: if you would like some help or advice, let me know!

cippy:
Hi,
I think the error concerning the message comes from having modified line 64 to have non-negative values as second argument of the round function, i.e.
absolute_digits = max(0,-value_precision + relative_digits)
I think I got the new line looking at another branch and assumed it was the best case
Now, I see the test is actually allowing for absolute_digits to be negative, which thus breaks the test.

I see other errors are related to style in naming variables or not using spaces after commas, will try to fix them and push again.

For the new functions, I think I should be able to devise a test in the 'tests/' folder

Marco

cippy:
I also noticed an error message during the tests, stating "itertools has no 'izip' member". 
This arises from the fact that in Python 3 itertools does no longer have the izip member, and the built-in zip should be used, which does the same job as itertools.izip was doing in older verisions.
However, I was using the code from lxplus where python 2.7 is available.

A possible workaround I found online could be to just use zip, but import itertools as in the following
"
try:
    from itertools import izip as zip # work in python < 3
except ImportError: 
    pass #  python is >= 3.x, no need to import itertools
"

clelange:
> I also noticed an error message during the tests, stating "itertools has no 'izip' member".
> This arises from the fact that in Python 3 itertools does no longer have the izip member, and the built-in zip should be used, which does the same job as itertools.izip was doing in older verisions.
> However, I was using the code from lxplus where python 2.7 is available.
> 
> A possible workaround I found online could be to just use zip, but import itertools as in the following
> "
> try:
> from itertools import izip as zip # work in python < 3
> except ImportError:
> pass # python is >= 3.x, no need to import itertools
> "

I think this approach is reasonable. We do something similar at https://github.com/HEPData/hepdata_lib/blob/master/hepdata_lib/__init__.py#L17-L21.

However, we now have [six](https://six.readthedocs.io/) as a requirement for the tests (see https://github.com/HEPData/hepdata_lib/blob/master/setup.py#L40-L41), and I think there's no harm actually making it a full dependency. You could then work around this issue by simply doing

```python
from six.moves import zip
```

Your suggestion is equally valid though and I have no strong opinion.

clelange:
@cippy Thanks a lot for your efforts! I made a few changes to `test_helpers.py` so that `pylint` doesn't complain anymore. The changes are made directly to your branch, so you can just fetch from your remote and merge (or simply pull).

hepdata_lib/helpers.py
```diff
@@ -58,13 +59,139 @@ def relative_round(value, relative_digits):
...
+    relative precision of first argument with respect to the second one
+    value and reference must be of the same type
+    usually they are float, but should also work for tuples
+
+    : param value: first value
+    : type  value: float
+
+    : param reference: reference value (usually the uncertainty on value)
+    : type  reference: float
+    """
```
clelange, line 90: Since you write "value and reference must be of the same type", I would suggest you actually check this here, and otherwise throw an exception (and you could then also add a test for that).

tests/test_helpers.py
```diff
@@ -30,3 +31,33 @@ def test_relative_round(self):
...
+        # uncertainty has two value, as it would be the case with TGraphAsymmErrors
+        cont_asymm_err = {"val": [1.23456, 0.123],
+                          "unc": [(0.00123, 0.0123), (0.012, 0.12)],
+                          "val_round": [1.235, 0.12],
+                          "unc_round": [(0.001, 0.012), (0.01, 0.12)]}
+        # round to two significant digits
+        round_value_and_uncertainty(cont_asymm_err, "val", "unc", 2)
+        for index in range(len(cont_asymm_err["val"])):
+            self.assertTrue(cont_asymm_err["val"] == cont_asymm_err["val_round"])
+            self.assertTrue(cont_asymm_err["unc"] == cont_asymm_err["unc_round"])
```
clelange, line 63: Should these lines actually use `[index]` to check each individual item? I changed that with my last commit in lines 49-50, but wanted to ask what was your intention here.

hepdata_lib/helpers.py
cippy: Yes, that's a good suggestion, I should definitely have made that. 
I can add those tests, no problem, but I think I won't manage to do it before the weekend

tests/test_helpers.py
cippy: Your are right, initially I was thinking of making a check item by item, but eventually I forgot. However, I think the direct comparison of the arrays is just equivalent and even simpler, the loop could be removed

hepdata_lib/helpers.py
```diff
@@ -58,13 +59,139 @@ def relative_round(value, relative_digits):
     if isinstance(value, tuple):
         return (relative_round(x, relative_digits) for x in value)
 
-    precision = math.ceil(math.log10(abs(value)))
-
-    absolute_digits = - precision + relative_digits
+    value_precision = math.ceil(math.log10(abs(value)))
```
AndreasAlbert, line 62: Would it make sense to replace this calculation with a call to `get_number_precision`?

cippy: Yes, now that I wrote that function it could be used here as well. Added in my ToDo list

hepdata_lib/helpers.py
clelange: Thanks! There's no rush, just take the time you need.

cippy:
Hi, sorry for the spam and the many commits to fix pylint errors.
Now it seems everything works fine except for tests/macospy3, but I have no idea why it happens. Perhaps you can better understand the origin of it.

With the last commits I implemented all the comments received above.
For the function get_value_precision_wrt_reference(), I eventually decide to accept only floats and ints as input arguments. I was initially implementing the case with both tuples or tuple (first argument) and float/int (for second one), but it required many more checks inside the function. Thus, since for the function was being used only to deal with floats, I just decided to go for the easier way. A user can deal with tuples by just using this simple function for each element, which seems easier than allowing for more complex behaviour inside the function itself.

I also added a fix to relative_round(), where I moved the check for the tuple input type as the first one. The problem was that if type(value) == tuple and the check is first made for 'value == 0', a ValueError is thrown, because 'The truth value of an array with more than one element is ambiguous'.
The same happens in get_number_precision() (there is also a comment to warn about this potential issue)

clelange:
Thanks for your work! I checked and can confirm that the failing test for MacOS is not due to your code, but rather other things going wrong with the pyROOT installation there.
From my side the code looks good to me. We should probably think about how we could document this in a more prominent way.
What do you think @AndreasAlbert ?

clelange:
@AndreasAlbert I'd merge this early next week unless you have additional comments. Let me know should you need more time.

AndreasAlbert:
@clelange @cippy thanks for the work! I don't have further comments. Apologies for holding this up, I missed the mention.

Issue #147, opened 2020-10-12
Deprecated ROOT.Long and ROOT.Double in ROOT 6.22


lucasflores:
Hello,

So I have been seeing this warning about ROOT.Double being deprecated in ROOT 6.22 for a while and successfully ignored it until now. In our latest software releases on ATLAS  we have finally moved to ROOT 6.22 and so these lines are finally broken https://github.com/HEPData/hepdata_lib/blob/master/hepdata_lib/root_utils.py#L410
The notes on this change in ROOT can be found here https://root.cern/doc/v622/release-notes.html#pyroot (or just search for ROOT.Double).
I was able to implement a fix in my version with the following changes in `root_utils.py` : 

```python
+ import ctypes
```
```python
L410| - x_val = r.Double()   L410| +  x_val = ctypes.c_double()
L411| - y_val = r.Double()   L411| +  y_val = ctypes.c_double()
```
 ```python
L413| - points["x"].append(float(x_val))   L413| + points["x"].append(float(x_val.value))
L414| - points["y"].append(float(x_val))   L414| + points["y"].append(float(x_val.value))
```

I can't say I have exhaustively tested this but doing a diff of  the `output/` HEPData directories (one using ROOT 6.20 with hepdatalib as is and one using ROOT 6.22 with my fix)  for my rather large and varied HEPData submission shows that their output is identical. 
I am not sure I have time these days to validate and merge this in in the nominal way myself but I thought I would bring it to your attention.

Best,
Lucas

clelange:
Hi Lucas,

Thanks for pointing this out and providing a solution! We should soon have some personpower to help with the maintenance of hepdata_lib to get this implemented and validated. Do you happen to know if this is backward-compatible for all ROOT 6 versions?

Clemens

lucasflores:
Hi Clemens,

Thanks for getting back tome, and yeah no problem. 

> Do you happen to know if this is backward-compatible for all ROOT 6 versions?

hmm I would not really know this :/  sorry, and I have not tested it on any earlier version than 6.20. However `ctypes` seems to be a pretty standard python library so I would think it should. It seems ROOT was just trying to move away from their own wheel inventions. 

And if your person power needs a next thing to do :) I also have in my fork here https://github.com/lucasflores/hepdata_lib implemented `THStack` and `THGraph2D (THGraph2DErrors)` functionality. I might have some more time soon myself to contribute more officially but I just thought I would share.

Best,
Lucas

[...]

PR #149, opened 2021-01-13
ROOT.Double --> ctypes

emanueleusai:
Tested on ROOT 6.22, 6.20, and 6.18 using LCG_96bpython3, LCG_97apython3, and LCG_98python3, respectively (x86_64-centos7-gcc8 for all three).

### DIFF
--- a/hepdata_lib/root_utils.py
+++ b/hepdata_lib/root_utils.py
@@ -3,6 +3,7 @@
 from collections import defaultdict
 import numpy as np
 import ROOT as r
+import ctypes
 from hepdata_lib.helpers import check_file_existence
 
 class RootFileReader(object):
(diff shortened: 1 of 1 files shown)


AndreasAlbert:
Hi @emanueleusai, thank you for preparing this PR! It seems that there is an unrelated issue with the CI checks failing that we need to look into. I ran the tests manually on my machine and see that the tests related to root file reading are fine, so we are good to go here.

Closes #147

clelange:
There's on linter error: 

```
************* Module hepdata_lib.root_utils
C:  6, 0: standard import "import ctypes" should be placed before "import numpy as np" (wrong-import-order)
```

The rest seem to be issues with the runners, looks like they changed the base images or so....

AndreasAlbert:
Well spotted, I missed that! Addressed in #150 together with other testing fixes.

emanueleusai:
Thanks for spotting the linter error on the import order. I completely missed that among the other CI errors unrelated to my PR.

PR #156, opened 2021-03-17
Avoid casting int to float unnecessarily

AndreasAlbert:
Our current input handling converts Variable inputs to float if they are integers. This is not needed and is detrimental when one wants to converse the discreteness of the input, e.g. for data counts. In this PR, I clean up the handling of the input values and avoid the unneeded cast.


AndreasAlbert:
Let's solve #158 as well here while we are at it

hepdata_lib/__init__.py
```diff
@@ -84,11 +84,11 @@ def values(self, value_list):
                                 )
 
             # All good
-            self._values = [(float(x[0]), float(x[1])) for x in value_list]
+            self._values = [(helpers.sanitize_value(x[0]), helpers.sanitize_value(x[1])) for x in value_list]
         else:
             # Check that the input is well-formed
             try:
-                parsed_values = [x if isinstance(x, str) else float(x) for x in value_list]
+                parsed_values = [helpers.sanitize_value(x) for x in  value_list]
```
clelange, line 91: Remove additional space after `in`

hepdata_lib/helpers.py
```diff
@@ -246,3 +246,19 @@ def check_file_size(path_to_file, upper_limit=None, lower_limit=None):
...
+    :type value: string, int, or castable to float
+
+    Strings and integers are left alone,
+    everything else is converted to float.
+    """
+    if isinstance(value,str):
+        return value
+    if isinstance(value,int):
+        return value
+    return float(value)
```
clelange, line 264: Add new line add end of file

hepdata_lib/__init__.py
```diff
@@ -84,11 +84,11 @@ def values(self, value_list):
...
             except (TypeError, ValueError):
```
clelange, line 92: Not sure it makes sense, but should we then raise the error in the helper as well (i.e. move the try/catch part over)?

clelange (changes requested):
Thanks for this! It's a bit unfortunate that our tests all fail at the moment. Did you happen to run them locally?

hepdata_lib/__init__.py
AndreasAlbert: done

hepdata_lib/helpers.py
AndreasAlbert: done

hepdata_lib/__init__.py
AndreasAlbert: If the exception happens within the helper, it will be on a call to `float(x)`, and the resulting error message will be very clear. I think this is fine the way it is.

AndreasAlbert:
Closes #158

AndreasAlbert:
Confirm that tests run locally (still have to fix CI)

clelange:
Also this one would have to be rebased and conflicts to be resolved...

AndreasAlbert:
Not sure why coveralls rates this as a failure. It seems all individual files either have unchanged or improved coverage, but the overall coverage reduced by some tiny amount. Rounding error?

image

AndreasAlbert:
I changed the settings on coveralls.io so that it will only fail for coverage decreases of 2% or more in the future. For this PR, suggest to ignore and merge @clelange

clelange (approved):
Awesome, thanks! I agree that we should ignore coverage here 😄

PR #161, opened 2021-04-19
Do not write errors if all errors are zero for bin

### LINKED ISSUE #159: Check that uncertainties are not all zero
In February 2020, I modified the [hepdata-validator](https://github.com/HEPData/hepdata-validator) code (see HEPData/hepdata-validator#7) to invalidate bins where all uncertainties are zero, in response to [problems](https://gitlab.com/hepcedar/rivet/-/issues/69) raised by the Rivet developers in fitting applications.  I've just added a paragraph of explanation to the [Uncertainties](https://hepdata-submission.readthedocs.io/en/latest/data_yaml.html#uncertainties) section of the submission documentation.  A similar check should be made in the `hepdata_lib` code, perhaps before the [lines](https://github.com/HEPData/hepdata_lib/blob/f7f85a9b8b86ef86f4b04fd373a7bb6cd530504b/hepdata_lib/__init__.py#L176-L193) constructing the `errors` key.  If all uncertainties are zero, the errors key could be omitted from the dictionary or set to an empty list, and a warning message should be printed out.  This issue was prompted by a [question](https://hepdata-forum.cern.ch/t/uncertainties-should-not-all-be-zero-error/22/3?u=graemewatt) in the [HEPData Forum](https://hepdata-forum.cern.ch).

AndreasAlbert:
Closes #159 

@GraemeWatt Is this what you imagined?


clelange:
Can you please rebase onto master to have working tests again?

hepdata_lib/helpers.py
```diff
@@ -246,3 +246,17 @@ def check_file_size(path_to_file, upper_limit=None, lower_limit=None):
...
+    """
+    Return a mask of bins where any of the uncertainties is nonzero.
+    """
+    nonzero = np.zeros(size, dtype=bool)
+
+    for unc in uncertainties:
+        if unc.is_symmetric:
+            nonzero = nonzero | (np.array(unc.values) > 0)
+        else:
+            nonzero = nonzero | np.any((np.array(unc.values) > 0),axis=1)
```
GraemeWatt, line 261: Here, I'd replace `> 0` with `!= 0` in lines 259 and 161 to allow for possible uncertainties with a flipped sign (see first paragraph of the [Uncertainties](https://hepdata-submission.readthedocs.io/en/latest/data_yaml.html#uncertainties) section in the submission docs).

hepdata_lib/__init__.py
```diff
@@ -173,25 +177,27 @@ def make_dict(self):
...
-                        "asymerror": {
-                            "minus":
-                                helpers.relative_round(unc.values[i][0], self.digits),
-                            "plus":
-                                helpers.relative_round(unc.values[i][1], self.digits)
-                        },
-                        "label": unc.label
-                    })
+            # An uncertainty entry is only appended
+            # if at least one of the uncertanities is not zero.
```
GraemeWatt, line 181: Typo: "uncertanities" → "uncertainties".

hepdata_lib/__init__.py
```diff
@@ -173,25 +177,27 @@ def make_dict(self):
...
+                    else:
+                        valuedict['errors'].append({
+                            "asymerror": {
+                                "minus":
+                                    helpers.relative_round(unc.values[i][0], self.digits),
+                                "plus":
+                                    helpers.relative_round(unc.values[i][1], self.digits)
+                            },
+                            "label": unc.label
+                        })
```
GraemeWatt, line 200: The omission of the `errors` key should not be done silently, but a warning message should be printed out.  Otherwise, users will encode bins with zero uncertainties (although `errors` will now be suppressed), which will still result in a YODA export with zero uncertainties that was the source of the [Rivet/Contur problems](https://gitlab.com/hepcedar/rivet/-/issues/69).  I suggest adding an `elif` condition like the following:
```python
            if nonzero_uncs[i]:
                [...]
            elif self.uncertainties:
                print("Warning: omitting 'errors' since all uncertainties are zero for bin {} of variable '{}'.".format(i, self.name))
                print("Note that bins with zero content should preferably be omitted completely from the HEPData table.")
```

GraemeWatt (changes requested):
Many thanks for addressing the issue #159 so quickly!  I've tested the new code and it works well, but I've made a couple of suggestions for improvements before merging.

GraemeWatt:
@AndreasAlbert, just to mention this issue was raised today by an ATLAS uploader, so it would be good to soon have the PR merged after addressing my comments.

hepdata_lib/__init__.py
AndreasAlbert: Change implemented as proposed

hepdata_lib/__init__.py
AndreasAlbert: Change implemented as proposed

hepdata_lib/helpers.py
AndreasAlbert: Change implemented as proposed

AndreasAlbert:
@GraemeWatt thanks for reminding me. I implemented your proposals.

AndreasAlbert:
It's slightly annoying that pylint requirements keep changing under our feet. We should consider fixing pylint to a given version to avoid this (would however require us to periodically upgrade the version and clean up the code according to new requirements then).

AndreasAlbert:
@GraemeWatt ok, good to merge now if you are happy with the changes

PR #164, opened 2021-05-06
Dockerfile: Make sure to use Fedora-based root container

AndreasAlbert:
The setup in the Dockerfile is based on the assumption that the base container is Fedora based. However, apparently root changed their tag structure, so 'latest' does not point to Fedora anymore. They still have specific Fedora tags and switching to that fixes the Docker build. @clelange please check if this makes sense to you

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rootproject/root:latest
+FROM rootproject/root:6.22.06-fedora33
 
 ENV MAGICK_HOME=/usr
 


AndreasAlbert:
I testes this by running:
```
docker build --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` --build-arg VCS_REF=`git rev-parse --short HEAD` --build-arg VCS_URL=`git config --get remote.origin.url`  .
```
which produced a successful build.

clelange (approved):
Awesome, thanks a lot! Shall we create new release once this is merged?

PR #168, opened 2021-06-14
RootFileReader: Allow access to bin labels + small other improvements

AndreasAlbert:
The main change here is the addition of the ability to read bin labels from ROOT histogram. Tests of the functionality have been added, as well.
While implementing the tests, I realized that many test functions have unnecessarily strict checks of the reader output, which break if you add new keys. I took the opportunity to change these assertions to only check that the data needed for a specific check is present, rather than enforcing that no additional data is present.

I also made two small quality-of-life changes to `Table`.

### DIFF
--- a/hepdata_lib/root_utils.py
+++ b/hepdata_lib/root_utils.py
@@ -284,7 +284,7 @@ def get_hist_2d_points(hist, **kwargs):
         assert ylim[0] < ylim[1]
 
     points = {}
-    for key in ["x", "y", "x_edges", "y_edges", "z", "dz","x_labels","y_labels"]:
+    for key in ["x", "y", "x_edges", "y_edges", "z", "dz", "x_labels", "y_labels"]:
         points[key] = []
 
     ixmin = hist.GetXaxis().FindBin(xlim[0]) if xlim[0] is not None else 1
--- a/hepdata_lib/root_utils.py
+++ b/hepdata_lib/root_utils.py
@@ -284,7 +284,7 @@ def get_hist_2d_points(hist, **kwargs):
         assert ylim[0] < ylim[1]
 
     points = {}
-    for key in ["x", "y", "x_edges", "y_edges", "z", "dz"]:
+    for key in ["x", "y", "x_edges", "y_edges", "z", "dz","x_labels","y_labels"]:
         points[key] = []
 
     ixmin = hist.GetXaxis().FindBin(xlim[0]) if xlim[0] is not None else 1
@@ -317,6 +317,9 @@ def get_hist_2d_points(hist, **kwargs):
 
             points["z"].append(z_val)
             points["dz"].append(dz_val)
+
+            points["x_labels"].append(hist.GetXaxis().GetBinLabel(x_bin))
+            points["y_labels"].append(hist.GetYaxis().GetBinLabel(y_bin))
 
     return points
 
@@ -356,7 +359,7 @@ def get_hist_1d_points(hist, **kwargs):
         assert xlim[0] < xlim[1]
 
     points = {}
-    for key in ["x", "y", "x_edges", "dy"]:
+    for key in ["x", "y", "x_edges", "x_labels", "dy"]:
         points[key] = []
 
     symmetric = (hist.GetBinErrorOption() == r.TH1.kNormal)
@@ -379,6 +382,8 @@ def get_hist_1d_points(hist, **kwargs):
 
         points["y"].append(y_val)
         points["dy"].append(dy_val)
+
+        points["x_labels"].append(hist.GetXaxis().GetBinLabel(x_bin))
 
     return points
 
--- a/tests/test_rootfilereader.py
+++ b/tests/test_rootfilereader.py
@@ -388,7 +388,7 @@ class TestRootFileReader(TestCase):
 
         # Clean up
         self.doCleanups()
-    
+
     def test_read_hist_2d_bin_labels(self):
         """Test the read_hist_2d function with bin labels."""
         name = "test"
--- a/tests/test_rootfilereader.py
+++ b/tests/test_rootfilereader.py
@@ -178,11 +178,15 @@ class TestRootFileReader(TestCase):
         x_values = [0.5 + x for x in range(N)]
         y_values = list(np.random.uniform(-1e3, 1e3, N))
         dy_values = list(np.random.uniform(0, 1e3, N))
-
+        x_edges = []
         hist = ROOT.TH1D("test1d_symm", "test1d_symm", N, 0, N)
         for i in range(1, hist.GetNbinsX()+1):
             hist.SetBinContent(i, y_values[i-1])
             hist.SetBinError(i, dy_values[i-1])
+
+            c = hist.GetBinCenter(i)
+            w = hist.GetBinWidth(i)
+            x_edges.append((c-0.5*w, c+0.5*w))
 
         testfile = make_tmp_root_file(testcase=self)
         testfile.cd()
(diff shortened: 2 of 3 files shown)


hepdata_lib/root_utils.py
```diff
@@ -284,7 +284,7 @@ def get_hist_2d_points(hist, **kwargs):
         assert ylim[0] < ylim[1]
 
     points = {}
-    for key in ["x", "y", "x_edges", "y_edges", "z", "dz"]:
+    for key in ["x", "y", "x_edges", "y_edges", "z", "dz","x_labels","y_labels"]:
```
clelange, line 287: Please add a whitespace after the comma :-)

tests/test_rootfilereader.py
```diff
@@ -114,7 +114,9 @@ def test_read_graph_tgrapherrors(self):
         reader = RootFileReader(testfile.GetName())
         data = reader.read_graph(name)
 
-        self.assertTrue(set(data.keys()) == set(["x", "y", "dx", "dy"]))
+        # Check consistency
+        for key in ["x", "y", "dx","dy"]:
```
clelange, line 118: Also here a whitespace would be nice

tests/test_rootfilereader.py
```diff
@@ -154,7 +156,10 @@ def test_read_graph_tgrapherrors2(self):
...
```
clelange, line 160: Whitespace

tests/test_rootfilereader.py
```diff
@@ -188,7 +193,9 @@ def test_read_hist_1d_symmetric_errors(self):
         reader = RootFileReader(testfile.GetName())
         points = reader.read_hist_1d(name)
 
-        self.assertTrue(set(["x", "y", "x_edges", "dy"]) == set(points.keys()))
+        # Check consistency
+        for key in ["x", "y", "dy"]:
```
clelange, line 197: Shouldn't you also check for `"x_edges"` here?

tests/test_rootfilereader.py
```diff
@@ -249,7 +256,9 @@ def test_read_hist_1d_range(self):
 
         points = reader.read_hist_1d(testname, xlim=(xmin, xmax))
 
-        self.assertTrue(set(["x", "y", "x_edges", "dy"]) == set(points.keys()))
+        # Check consistency
+        for key in ["x", "y", "dy"]:
```
clelange, line 260: As above: shouldn't you also check for `"x_edges"` here?

tests/test_rootfilereader.py
```diff
@@ -337,7 +347,8 @@ def test_read_hist_1d_asymmetric_force_symmetric_errors(self):
         points = reader.read_hist_1d(testname, force_symmetric_errors=True)
 
         # Check consistency
-        self.assertTrue(set(["x", "y", "x_edges", "dy"]) == set(points.keys()))
+        for key in ["x", "y", "dy"]:
```
clelange, line 350: As above: shouldn't you also check for `"x_edges"` here?

tests/test_rootfilereader.py
```diff
@@ -347,11 +358,80 @@ def test_read_hist_1d_asymmetric_force_symmetric_errors(self):
...
+        y_values = np.random.uniform(-1e3, 1e3, (N,N))
+        xlabels = ["X{0}".format(i) for i in range(N)]
+        ylabels = ["Y{0}".format(i) for i in range(N)]
+
+        hist = ROOT.TH2D("test2d_labels", "test2d_labels", N, 0, N, N, 0, N)
+        for i in range(hist.GetNbinsX()):
+            for j in range(hist.GetNbinsY()):
+                hist.Fill(N,N,y_values[i,j])
+                hist.GetXaxis().SetBinLabel(i+1, xlabels[i])
+                hist.GetYaxis().SetBinLabel(j+1, ylabels[i])
```
clelange, line 407: `ylabels[i]` -> `ylabels[j]`? This could have been caught had the two axes different lengths (but only if x > y).

tests/test_rootfilereader.py
```diff
@@ -347,11 +358,80 @@ def test_read_hist_1d_asymmetric_force_symmetric_errors(self):
...
+        points = reader.read_hist_2d(name)
+
+        self.assertTrue("x_labels" in points.keys())
+        self.assertTrue("y_labels" in points.keys())
+
+        # The output ordering is
+        # [(x=0,y=0), (x=0,y=1), ...]
+        for i in range(N*N):
+            self.assertTrue(
+                points["x_labels"][i]==xlabels[i//N]
```
clelange, line 424: Maybe add a comment that `//` is the integer division operator. Still, I think an example with x bins != y bins would be better.

tests/test_rootfilereader.py
```diff
@@ -382,7 +462,8 @@ def test_read_hist_2d_symmetric_errors(self):
         points = reader.read_hist_2d(testname)
 
         # Check keys
-        self.assertTrue(set(["x", "y", "z", "x_edges", "y_edges", "dz"]) == set(points.keys()))
+        for key in ["x","y","z","dz"]:
```
clelange, line 465: spaces? :-)
Also, as above, what about the edges?

tests/test_rootfilereader.py
```diff
@@ -470,7 +551,8 @@ def test_read_hist_2d_range(self):
         points = reader.read_hist_2d(testname, xlim=(xmin, xmax), ylim=(ymin, ymax))
 
         # Check keys
-        self.assertTrue(set(["x", "y", "z", "x_edges", "y_edges", "dz"]) == set(points.keys()))
+        for key in ["x","y","z","dz"]:
```
clelange, line 554: See above

tests/test_rootfilereader.py
```diff
@@ -525,7 +607,8 @@ def test_read_hist_2d_asymmetric_errors(self):
         points = reader.read_hist_2d("test2d_asym")
 
         # Check keys
-        self.assertTrue(set(["x", "y", "z", "x_edges", "y_edges", "dz"]) == set(points.keys()))
+        for key in ["x","y","z","dz"]:
```
clelange, line 610: See above, spaces and edges.

tests/test_rootfilereader.py
```diff
@@ -578,7 +661,8 @@ def test_read_hist_2d_asymmetric_force_symmetric_errors(self):
         points = reader.read_hist_2d("test2d_asym", force_symmetric_errors=True)
 
         # Check keys
-        self.assertTrue(set(["x", "y", "z", "x_edges", "y_edges", "dz"]) == set(points.keys()))
+        for key in ["x","y","z","dz"]:
```
clelange, line 664: See above, spaces and edges.

clelange (changes requested):
I left a few comments in the code.

I don't understand why you dropped the `edges` from the tests.

For the 2D bin label test, I'd suggest to use different numbers of bins for x and y.

tests/test_rootfilereader.py
AndreasAlbert: My logic here is to only check for the keys that are explicitly tested in the coming lines. The actual testing in this function does not care about the edges. If we want to check the presence of edges, we should also test their values.

AndreasAlbert:
I added an actual test of the values of the edges now. (Still following the idea that the presence of a key is only tested in a place where the values of said key are tested, too)

tests/test_rootfilereader.py
AndreasAlbert: Slightly rewrote this piece

AndreasAlbert:
I think I got all the white space, too

clelange:
Awesome, looks good to me! Glad this is improving your quality of life.

clelange (approved):

Issue #35, opened 2018-06-11
Add reference to hepdata_lib in additional_resources


AndreasAlbert:

clelange:
Add `additional_resources` field to submission so that we can figure out that this library has been used. Would be useful to also include the version number. Hopefully, the search interface allows to filter.

PR #170, opened 2021-06-16
Factor out additional resources, add to Table

### LINKED ISSUE #157: Add "add_additional_resource" method to Table class
I finally got around to trying out the `hepdata_lib` package.  It seems quite easy to use after I got ROOT installed on my laptop (macOS), so thanks for providing such a helpful tool!  An option to install without ROOT (issue #99) would indeed be useful if working only with plain-text files as input.  I had a use case where I wanted to write event counts as integers not floats, so I used the branch from PR #156.  I've added my two scripts using `hepdata_lib` as new examples to the [HEPData submission documentation](https://hepdata-submission.readthedocs.io/en/latest/examples.html) in a commit https://github.com/HEPData/hepdata-submission/commit/28d290c5f880a507998e22497efca7cb116daa91.

One missing feature I noticed is an `add_additional_resource` method for the `Table` class similar to the existing method of the same name for the `Submission` class.  I was able to work around this limitation by doing something like:
```python
table.additional_resources.append({"description": description, "location": filename})
submission.files_to_copy.append(location)
```

But it would be more convenient to be able to simply do:
```python
table.add_additional_resource(description, location, copy_file=True)
```
as for resources attached to the whole `Submission` object.

AndreasAlbert:
Functionality related to additional resources has now been separated out into a mixin that is flexible inherited into both Table and Submission. When the Submission object creates the output files, it automatically calls the copy_files method of all tables.

Closes #157

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -83,11 +83,11 @@ class AdditionalResourceMixin(object):
             helpers.check_file_existence(location)
             helpers.check_file_size(location, upper_limit=100)
             resource["location"] = os.path.basename(location)
+            self.files_to_copy.append(location)
         else:
             resource["location"] = location
 
         self.additional_resources.append(resource)
-        self.files_to_copy.append(location)
 
     def copy_files(self, outdir):
         """
@@ -450,6 +450,7 @@ class Submission(AdditionalResourceMixin):
         self.tables = []
         self.comment = ""
         self.record_ids = []
+        self.add_additional_resource("Created with hepdata_lib " + __version__, "https://zenodo.org/record/4946277")
 
     @staticmethod
     def get_license():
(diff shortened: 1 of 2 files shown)


AndreasAlbert:
Closes #35

hepdata_lib/__init__.py
```diff
@@ -385,19 +440,21 @@ def write_yaml(self, outdir="."):
...
     def __init__(self):
+        super().__init__()
         self.tables = []
         self.comment = ""
-        self.additional_resources = []
         self.record_ids = []
-        self.files_to_copy = []
+        self.add_additional_resource(
+            "Created with hepdata_lib " + __version__,
+            "https://zenodo.org/record/4946277")
```
clelange, line 457: Nice test of this :-)

clelange (approved):
Thanks for this nice piece of work! Looks good to me!

PR #174, opened 2021-08-19
Fix missing file in tar ball

### LINKED ISSUE #173: Additional resources not copied to .tar.gz when added through Table
Hey,

when adding additional resources through Table rather than Submission object the additional resources are not being added to the submission tarball created.

The behaviour can be reproduced with the following code:
```
import hepdata_lib
sub=hepdata_lib.Submission()
tab=hepdata_lib.Table('test')
tab.add_additional_resource("replica1","example/ph1_pt_replicas.tar.gz",True)
sub.add_table(tab)
sub.create_files()
```
the resulting tarball is missing `ph1_pt_replicas.tar.gz`, however untared directory with submission files contains the required object fine. 

The bug can be traced to https://github.com/HEPData/hepdata_lib/blob/master/hepdata_lib/__init__.py#L562-L563 where only files in Submission object's 'files_to_copy' are copied to the tarball. 'files_to_copy' inside tables are not looped over.

AndreasAlbert:
Fixes #173 

Summary of changes:
* Slight rework of how Submission handles file copying and tar ball creation. Now correctly propagates files from its Table daughters. Added explicit test for this case.
* Small changes forced by new pylint version. Have fixed pylint version now so that it does not keep changing randomly.
* While cleaning up pylint, encountered and fixed small bug in helpers where two functions returned generator objects rather than tuples.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ try:
 except ImportError:
     print("ROOT is required by this library.")
 
-DEPS = ['numpy', 'PyYAML>4.*', 'future', 'pylint']
+DEPS = ['numpy', 'PyYAML>4.*', 'future', 'pylint==2.9.6']
 
 HERE = path.abspath(path.dirname(__file__))
 
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -226,16 +226,16 @@ class Variable(object):
                                                         self.uncertainties,
                                                         size=len(self._values)
                                                         )
-        for i in range(len(self._values)):
+        for i, value in enumerate(self._values):
             valuedict = defaultdict(list)
 
             if self.is_binned:
-                valuedict["low"] = helpers.relative_round(self._values[i][0],
+                valuedict["low"] = helpers.relative_round(value[0],
                                                           self.digits)
-                valuedict["high"] = helpers.relative_round(self._values[i][1],
+                valuedict["high"] = helpers.relative_round(value[1],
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -8,7 +8,7 @@
 except ImportError:
     print("ROOT is required by this library.")
 
-DEPS = ['numpy', 'PyYAML>4.*', 'future', 'pylint']
+DEPS = ['numpy', 'PyYAML>4.*', 'future', 'pylint==2.9.6']
```
clelange, line 11: Just wondering whether it wouldn't be better to not fix the `pylint` version and rather pay the price of having to fix the code in some places from time to time with the benefit of improved code quality.

clelange (commented):
Hey, I've been on vacation. This looks good to me, just wondering about fixing the `pylint` version, see inline comment.

setup.py
AndreasAlbert: I think its more efficient in terms of person power to decouple the pylint updates from regular PRs. Based on the experience of the last few months here, I have encountered new unexpected pylint complaints with every PR I've made. This means that a) every PR touches code all over the repository and b) we waste time by having extra iterations of failing tests -> local fix -> push -> test again. I do not think that that annoyance is worth the very minor improvements we get from cutting edge pylint. We can incorporate these any time by making a separate PR that deals with nothing but an update in pylint version and makes the required code style changes.

clelange (approved):
LGTM

PR #176, opened 2021-08-25
Handle nested TPads to get TObject

longjon929:
I've added another attempt at retrieving the TObject from nested TPads if the path fails.  This is useful in the case of a TCanvas with TPads for ratio panels etc.

### DIFF
--- a/hepdata_lib/root_utils.py
+++ b/hepdata_lib/root_utils.py
@@ -90,10 +90,27 @@ class RootFileReader(object):
             assert False
 
         except AssertionError as err:
-            msg="Cannot find any object in file {0} with path {1}".format(
+            msg="Cannot find any object in file {0} with path {1}.  Will try nested TPads".format(
                     self.tfile, path_to_object)
+            print(msg)
+        # If the above didn't work, try nested tpads
+        try:
+            obj = self.tfile.Get(parts[0])
+            for part in parts[1:]:
+                obj = obj.GetPrimitive(part)
+            
+            assert obj
+
+            return obj
+        except AssertionError as err:
+            msg="Cannot find any nested TPads in file {0} with path {1}".format(
+                    self.tfile, path_to_object)
+
             raise_from(IOError(msg), err)
+
         return None
+
+
 
     def read_graph(self, path_to_graph):
         """Extract lists of X and Y values from a TGraph.
--- a/hepdata_lib/root_utils.py
+++ b/hepdata_lib/root_utils.py
@@ -52,12 +52,15 @@ class RootFileReader(object):
         """
         Generalized function to retrieve a TObject from a file.
 
-        There are two use cases:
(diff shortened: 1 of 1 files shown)


AndreasAlbert:
Thanks for the feature implementation! Would you be able to also add a few simple tests? (like e.g. [here](https://github.com/HEPData/hepdata_lib/blob/master/tests/test_rootfilereader.py))

longjon929:
Hi @AndreasAlbert , sure.  I've tried copying a similar function and modifying it to test this case.  It does print out a few lines like this 
"Cannot find any object in file Name: tmp_7R6HNP4ITWDS.root Title:  with path canvas/pad1/testhist.  Will try nested TPads
.Cannot find any object in file Name: tmp_2TMNFIAEHX4Z.root Title:  with path Some/Nonsense/Path.  Will try nested TPads"
because the first try no longer raises and error.  If you would prefer, I can change it to only print something if both attempts fail.

AndreasAlbert:
thanks! yes, indeed, it would be optimal to only print warnings in case all strategies fail.

longjon929:
Ok, I've tidied up the messages.

clelange:
Hi @longjon929 - thanks for addressing the additional comments! The code looks good to me. Our automated tests, which also include running [pylint](https://www.pylint.org/) show some code formatting issues. You can see the details in the check details, I'm also adding them below:

```
hepdata_lib/__init__.py:229:8: C0200: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
hepdata_lib/__init__.py:417:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
hepdata_lib/__init__.py:422:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
hepdata_lib/__init__.py:512:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
hepdata_lib/__init__.py:541:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module hepdata_lib.c_file_reader
hepdata_lib/c_file_reader.py:38:30: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module hepdata_lib.root_utils
hepdata_lib/root_utils.py:104:0: C0303: Trailing whitespace (trailing-whitespace)
hepdata_lib/root_utils.py:109:0: C0301: Line too long (146/100) (line-too-long)
hepdata_lib/root_utils.py:113:0: C0301: Line too long (146/100) (line-too-long)
hepdata_lib/root_utils.py:95:8: W0612: Unused variable 'err' (unused-variable)
```

Let us know if you have troubles addressing them and we can help you.

longjon929:
Hi @clelange, I've pushed a commit that should fix the warnings in root_utils.py.  Since I didn't modify the other files, I didn't look into the other linting warnings.

PR #188, opened 2021-12-02
write_image: Only convert images if target is outdated.

### LINKED ISSUE #179: skip overwriting existing output when creating record
Hi,

Fantastic package, really made my life significantly easier when generating a pretty big record.  Thanks for supporting this!

One thing I noticed when debugging my record is that running `convert` to create PNG's and thumbnails is (not surprisingly) quite slow.  Often the convert commands are overwriting PNG's or thumbnails that have already been created, especially if we're iterating on relatively minor changes in the yaml files.  

It would be fantastic to have an option to not overwrite existing PNG files, or to skip PNG file creation altogether.  I can see how this isn't something that we'd want by default, but being able to pass in an option to `submission.create_files()` to skip PNG creation would be super useful for debugging.

Thanks again!

-Mike

AndreasAlbert:
Closes #179

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -358,26 +358,35 @@ class Table(AdditionalResourceMixin):
                 raise RuntimeError("File %s does not exist!" % image_file)
             if not os.path.exists(outdir):
                 os.makedirs(outdir)
-            out_image_file = "{}.png".format(
-                os.path.splitext(image_file)[0].rsplit("/", 1)[1])
-            thumb_out_image_file = "thumb_" + out_image_file
-            # first convert to png, then create thumbnail
-            command = "convert -flatten -density 300 -fuzz 1% -trim +repage {} {}/{}".format(
-                image_file, outdir, out_image_file)
-            command_ok = helpers.execute_command(command)
-            if not command_ok:
-                print("ImageMagick does not seem to be installed \
-                       or is not in the path - not adding any images.")
-                break
-            command = "convert -thumbnail 240x179 {outdir}/{image} {outdir}/{thumb}".format(
-                outdir=outdir, image=out_image_file, thumb=thumb_out_image_file)
-            helpers.execute_command(command)
+
+            # PNG file is named same as input file except for extension
+            # Thumbnail is named with a '_thumb' prefix
+            png_output_base = os.path.splitext(os.path.basename(image_file))[0] + ".png"
+            thumbnail_output_base = "thumb_" + png_output_base
+
+            # Absolute paths for further use
+            png_output_path = os.path.join(outdir, png_output_base)
+            thumbnail_output_path = os.path.join(outdir, thumbnail_output_base)
+
+
+            # Convert to full-size PNG image
+            # Only executed if output is missing or out of date
+            if helpers.file_is_outdated(png_output_path, image_file):
+                helpers.convert_pdf_to_png(image_file, png_output_path)
+            else:
+                print("Full-size PNG file %s is up to date." % image_file)
+
+            if helpers.file_is_outdated(thumbnail_output_path, png_output_path):
+                helpers.convert_png_to_thumbnail(png_output_path, thumbnail_output_path)
+            else:
+                print("Thumbnail PNG file %s is up to date." % thumbnail_output_path)
+
             image = {}
             image["description"] = "Image file"
-            image["location"] = out_image_file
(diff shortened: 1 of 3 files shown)


hepdata_lib/helpers.py
```diff
@@ -285,3 +285,58 @@ def sanitize_value(value):
...
+    :param file_path: Path to the file to check.
+    :type file_path: str
+    :param reference_file_path: Path to the reference file.
+    :type reference_file_path: str
+    """
+    if not os.path.exists(reference_file_path):
+        raise RuntimeError("Reference file does not exist: %s" % reference_file_path)
+    if not os.path.exists(file_path):
+        return True
+    return os.path.getmtime(file_path) < os.path.getmtime(reference_file_path)
```
clelange, line 342: I see a few cases where this could not give the desired behaviour, but it might also depend on the file system. Take e.g. that one make a mistake in naming the files. This would not alter the change date, so the PNGs would not be updated. I like the idea of being "smart" in general like here, but I think we need an option to force-replace/-update all images. What do you think?

clelange (changes requested):
Thanks a lot for implementing this!

I left comments in the code:

1. [ ] check if the return values for the conversion commands are needed
2. [ ] add a force-recreate option for PNG replacement

hepdata_lib/helpers.py
AndreasAlbert: Could you elaborate on your example case of incorrectly named files? How would that lead to a problem? The name of the converted file and thumbnail are derived from the original file, so if change the name of the input file, the previously created output file will not be checked anymore.

Regarding force-recreation: This is tricky to implement here because this function is under the surface and one would have to propagate the force-recreate option through multiple layers of functions. Instead, it might be easier to just add a `remove_old` option to `create_files` that ensures that the output directory is erased before rewriting. This will of course trigger a new image conversion. This option is implemented as part of #189

clelange: I hadn't had a look at #189, but this is a good option, so we can consider this part solved.

Regarding the example case, let's say you accidentally copy the figure from the preliminary result, then obtain the PNGs, and afterwards replace the figure with the one from the paper. The change date of the new figure will probably still be earlier than the one of the PNGs. It might depend on the file system/OS, at least for macOS the change date won't update when moving a file from one directory to another or copying using the Finder (while it does when copying using the `cp` command).

hepdata_lib/__init__.py
```diff
@@ -358,26 +358,35 @@ def write_images(self, outdir):
...
+
+            # Absolute paths for further use
+            png_output_path = os.path.join(outdir, png_output_base)
+            thumbnail_output_path = os.path.join(outdir, thumbnail_output_base)
+
+
+            # Convert to full-size PNG image
+            # Only executed if output is missing or out of date
+            if helpers.file_is_outdated(png_output_path, image_file):
+                helpers.convert_pdf_to_png(image_file, png_output_path)
```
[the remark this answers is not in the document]
AndreasAlbert: Removed the returns now.

hepdata_lib/__init__.py
```diff
@@ -358,26 +358,35 @@ def write_images(self, outdir):
...
+            else:
+                print("Full-size PNG file %s is up to date." % image_file)
+
+            if helpers.file_is_outdated(thumbnail_output_path, png_output_path):
+                helpers.convert_png_to_thumbnail(png_output_path, thumbnail_output_path)
```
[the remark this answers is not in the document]
AndreasAlbert: Removed the returns now.

hepdata_lib/helpers.py
clelange: I don't have a solution to this, maybe we change the output from `Full-size PNG file %s is up to date.` to `Will not recreate full-size PNG file %s since it is more recent than its source.` - not sure if we need an addition such as `Delete the PNG file or use remove_old=True in create_files to have it recreated`.

AndreasAlbert: I did not know that mtime did not change with moves! Thanks for pointing this out. I modified the code now s that the file is replaced if either the [modification time](https://docs.python.org/3/library/os.path.html#os.path.getmtime) or [change time](https://docs.python.org/3/library/os.path.html#os.path.getctime) is outdated. I believe this should as many edge cases as easily possible.

clelange (approved):
Thanks, looks good to me!

PR #189, opened 2021-12-02
Validation of output tarballs 

### LINKED ISSUE #160: Validate submission created by create_files method
After calling the `create_files` method of the `Submission` object, it would be good to validate the files created in the `outdir` directory using the [hepdata-validator](https://github.com/HEPData/hepdata-validator) package.  This could be done in a similar way to the [check.py](https://github.com/HEPData/hepdata-submission/blob/master/scripts/check.py) script, allowing mistakes to be caught offline before uploading to [hepdata.net](https://www.hepdata.net).  For example, the check that uncertainties are not all zero (see #159) would be caught by the validator.  The extra validation step should be switched on by default, perhaps with an option to switch it off.

AndreasAlbert:
With this PR, `Submission.create_files` now automatically calls the hepdata validator on the output file. An AssertionError is thrown if the tar ball is found to be invalid.

Closes #160

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,4 +2,3 @@ m2r
 sphinx_rtd_theme
 docutils<0.18
 mistune<2.0.0
-hepdata-validator>=0.3.2
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
 m2r
 sphinx_rtd_theme
 docutils<0.18
+hepdata-validator>=0.3.2
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -525,7 +525,7 @@ class Submission(AdditionalResourceMixin):
             files = files + table.files_to_copy
         return files
 
-    def create_files(self, outdir=".", validate=True):
+    def create_files(self, outdir=".", validate=True, remove_old=True):
         """
         Create the output files.
 
@@ -534,7 +534,12 @@ class Submission(AdditionalResourceMixin):
 
         If `validate` is True, the hepdata-validator package will be used to validate the
         output tar ball.
+
+        If `remove_old` is True, the output directory will be deleted before recreation.
         """
+        if remove_old and os.path.exists(outdir):
+            shutil.rmtree(outdir)
+
         if not os.path.exists(outdir):
             os.makedirs(outdir)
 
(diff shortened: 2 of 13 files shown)


hepdata_lib/__init__.py
```diff
@@ -567,13 +570,18 @@ def create_files(self, outdir="."):
...
             for filepath in files_to_add:
                 tar.add(
                         filepath,
                         arcname=os.path.basename(filepath)
                         )
 
+        if validate:
+            full_submission_validator = FullSubmissionValidator()
+            is_archive_valid = full_submission_validator.validate(archive=tarfile_path)
+            assert is_archive_valid, "The tar ball is not valid"
```
GraemeWatt, line 584: Thanks for implementing this!  I think you also need to print out the validator error messages so that the user gets a hint how to fix the invalid tarball, i.e. something like:
```python
if not is_archive_valid:
    for filename in full_submission_validator.get_messages():
        full_submission_validator.print_errors(filename)
```

AndreasAlbert: Thanks, good suggestion, included now.

AndreasAlbert:
I also added a new `remove_old` option to `create_files`. If set to True, the output directory is deleted before the new files are written. This ensures that no unrelated files accidentally end up in the tar ball, which the validator would identify as an error. Since many of our notebooks use the same output directory naming, consecutive execution of the notebooks -- as done in our tests -- can provoke this error if the folder is not removed between executions.

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 m2r
 sphinx_rtd_theme
-docutils<0.18
\ No newline at end of file
+docutils<0.18
+hepdata-validator>=0.3.2
```
clelange, line 4: Just a curiosity, but this is needed for the docs to build?

clelange (commented):
Lots of hidden gems in this PR ;-)

docs/requirements.txt
AndreasAlbert: Life inside VS code:
1. decide to edit requirements file
2. ctrl + p, type "require..."
3. spend no time checking whether it found the right file
4. edit, commit, done
:)

AndreasAlbert: removed now

clelange:
I forgot, shall we remove Python2 support with this PR? That would mean removing it from https://github.com/HEPData/hepdata_lib/blob/master/setup.py#L30

AndreasAlbert:
Removed python2 from setup.py now

clelange (approved):
Thanks, and there goes Python2!

PR #193, opened 2022-01-12
Submission.create_files: Set remove_old default to False

### LINKED ISSUE #192: submission.create_files() fails and wipes the directory clean
Hi!

this script based off the simple examples I found in the documentation, fails to write the submission file `submission.create_files()` and deletes all other files that reside in the same directory it is launched from.

```
import numpy as np
from hepdata_lib import Submission, Table, Variable
from gentau_pt_eff import raw_data # here https://gist.github.com/rmanzoni/801a8788a5c6b8db3b5d3cd6697097d5
  
submission = Submission()

table = Table("Figure 5")

tau_gen_vis_pt = Variable(
    "Efficiency times acceptance", 
    is_independent = True, 
    is_binned      = True, 
    units          = "GeV"
)
tau_gen_vis_pt.values = [ 
    ( 20., 22.), 
    ( 22., 24.), 
    ( 24., 26.), 
    ( 26., 28.), 
    ( 28., 30.), 
    ( 30., 35.), 
    ( 35., 40.), 
    ( 40., 50.), 
    ( 50., 60.), 
    ( 60., 70.), 
    ( 70., 80.), 
    ( 80., 90.),
    ( 90.,100.),
    (100.,150.),
    (150.,200.),
    (200.,500.),
]

table.add_variable(tau_gen_vis_pt)

efficiencies = dict()

nicer_names = {
    'dmw2p' :'decay mode',
    'dmwo2p':'decay mode w/o 2-prong',
    'loose' :'deeptau vs jet loose',
    'medium':'deeptau vs jet medium',
    'tight' :'deeptau vs jet tigt',
}

# dependent variables y-axis
for ik in raw_data.keys():
    efficiencies[ik] = Variable("Efficiency %s" %nicer_names[ik], is_independent = False, is_binned = False, units = "")
    efficiencies[ik].values = raw_data[ik]['values']
    table.add_variable(efficiencies[ik])

submission.add_table(table)
submission.create_files()

```
I guess this is not the intended behaviour...

this is the error it rreturns
```
In [2]: submission.create_files()
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-7ea7852cac01> in <module>
----> 1 submission.create_files()

~/opt/anaconda3/envs/hammer3p8/lib/python3.8/site-packages/hepdata_lib/__init__.py in create_files(self, outdir, validate, remove_old)
    552         """
    553         if remove_old and os.path.exists(outdir):
--> 554             shutil.rmtree(outdir)
    555
    556         if not os.path.exists(outdir):

~/opt/anaconda3/envs/hammer3p8/lib/python3.8/shutil.py in rmtree(path, ignore_errors, onerror)
    735             # can't continue even if onerror hook returns
    736             return
--> 737         return _rmtree_unsafe(path, onerror)
    738
    739 # Allow introspection of whether or not the hardening against symlink

~/opt/anaconda3/envs/hammer3p8/lib/python3.8/shutil.py in _rmtree_unsafe(path, onerror)
    617         os.rmdir(path)
    618     except OSError:
--> 619         onerror(os.rmdir, path, sys.exc_info())
    620
    621 # Version using fd-based APIs to protect against races

~/opt/anaconda3/envs/hammer3p8/lib/python3.8/shutil.py in _rmtree_unsafe(path, onerror)
    615                 onerror(os.unlink, fullname, sys.exc_info())
    616     try:
--> 617         os.rmdir(path)
    618     except OSError:
    619         onerror(os.rmdir, path, sys.exc_info())

OSError: [Errno 22] Invalid argument: '.'

In [3]:
```

AndreasAlbert:
Closes #192

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -538,7 +538,7 @@ class Submission(AdditionalResourceMixin):
             files = files + table.files_to_copy
         return files
 
-    def create_files(self, outdir=".", validate=True, remove_old=True):
+    def create_files(self, outdir=".", validate=True, remove_old=False):
         """
         Create the output files.
 


clelange:
Thanks for the quick fix! I think we should take this a bit further and actually prohibit having the directory, in which the python script resides, be the same as the output directory (and the script directory must also not be a subdirectory of the output directory). This should protect us from accidental file deletion. What do you think?

AndreasAlbert:
I support making this overall more robust. However, I think that tacking on individual checks for specific edge cases might not be the best approach as we might still miss something. How about instead, we generally force the user to use a previously nonexistant directory the first time they execute create_files. That would require us to preserve between runs the knowledge of whether or not a given directory was originally created by hepdata_lib. We could easily accomplish this by depositing an empty signifier file (e.g. `$DIRECTORY/.created_by_hepdata_lib`) in the desired directory. Each time it is run, `create_files` would check whether the output directory exists already and whether the signifier file exists.  If the directory does not yet exist, we proceed as normal with creating the directory as well as the signifier file. If the directory exists, but the file does not, we exit and give the user a warning telling them that they should use a dedicated empty directory in order to avoid trouble.

Right this moment, though, we have code published on pypi that can accidentally wipe user files with default settings. Therefore, let's please merge this hot fix and mint a new version. That buys us a little time to think through how to really fix this once and for all.

clelange (approved):

PR #201, opened 2022-10-30
Add dependabot config for GitHub actions

clelange:
Following instructions at https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

Later PRs could do the same for Python/pip and maybe also Docker.


clelange:
Will merge without review (also, not sure there's a reviewer left besides me). If things go crazy, we can remove the whole dependabot fun.

PR #213, opened 2022-11-09
Dealing with symmetric and asymmetric uncertainties at the same time

frgigr:
Hi! This MR adds more flexibility when dealing with distributions/plots having low non-zero yields: in these cases, due to the limited statistics, it may happen that the uncertainty is symmetric, but the relative value being larger with respect to the nominal value. Therefore, it's a common practice to cap the negative variation at the nominal value to avoid a negative yield, being unphysical.

From a technical point of view, uncertainties are given as asymmetric, but when the up/down variations have opposite sign only, thus sharing the same magnitude, they are considered as symmetric. On the other hand, nothing changes for asymmetric ones. So, the resulting _yaml_ file will have symmetric or asymmetric uncertainties accordingly, improving general readability in case of mixed entries. 

Finally, as uncertainties are floating-point numbers, the [decimal](https://docs.python.org/3/library/decimal.html) library is used to perform the mathematical comparison and calculation(s).

Hope it is fine but, please, let me know if you have comments and/or questions. 
Thanks in advance,
Francesco

----
:books: Documentation preview :books:: https://hepdata-lib--213.org.readthedocs.build/en/213/

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -8,6 +8,7 @@ import shutil
 import tarfile
 import warnings
 from collections import defaultdict
+from decimal import Decimal
 import numpy as np
 import yaml
 from future.utils import raise_from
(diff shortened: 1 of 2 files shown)


frgigr:
Hi @clelange,

Thanks for your comments! No problem, I can wait for you to get back here.

In the meanwhile, if the support for Python2 is still needed, a temporary solution (to avoid using the `decimal` class) would be to drop the line where the `sum_unc` variable is created and to replace the `if-clause` with the following one, although less robust:
```
if abs( float(unc.values[i][0]) ) == abs( float(unc.values[i][1]) )
```

frgigr:
Just added a dummy test, to cross-check the change in this MR: as it involves uncertainties, I added a new method in the dedicated file. 
Thanks!

clelange (approved):
I should just note that I was wrong and `Decimal` is actually also available in Python2, so there should not be any compatibility issues.

PR #218, opened 2023-02-21
fix: Use valid project/version requirement specifiers

matthewfeickert:
'install_requires' must be a string or list of strings containing valid project/version requirement specifiers. Without this fix installation will fail with modern pip.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy
-PyYAML>4.*
+PyYAML>=4.0
 future
 six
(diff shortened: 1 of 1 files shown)


matthewfeickert:
In general I would recommend against using a requirements.txt for install_requires. I don't think you get any advantage here and it leaves yourself open to various problems.

matthewfeickert:
@clelange this is ready for review.

clelange (approved):
Thanks, looks like a trivial (but important) change!

PR #243, opened 2023-10-27
Converting from scikit-hep histogram to Table

yimuchen:
- Adding the scikit-hep histograms conversion function 
- Adding help function to simplify the Table/Variable declaration
- Properly add test cases

----
:books: Documentation preview :books:: https://hepdata-lib--243.org.readthedocs.build/en/243/

### DIFF
--- a/tests/test_histread.py
+++ b/tests/test_histread.py
@@ -43,19 +43,55 @@ class TestHistUtils(TestCase):
     def test_default_read(self):
         """
         Ensure basic readout function generates arrays with compatible
-        dimensions with given the base-level histogram.
+        dimensions with given the base-level histogram. Also checking alternate
+        binning methods.
         """
-        try:
-            readout = read_hist(TestHistUtils.base_hist)
-        # pylint: disable=W0702
-        except:
-            self.fail("Histogram reading raised an unexpected exception.")
-        # pylint: enable=W0702
+        # Always test the base histogram
+        readout = read_hist(TestHistUtils.base_hist)
 
         # Checking dimension compatibility
-        self.assertTrue(len(readout["dataset"]) == len(readout["flavor"]))
-        self.assertTrue(len(readout["dataset"]) == len(readout["eta"]))
-        self.assertTrue(len(readout["dataset"]) == len(readout["pt"]))
+        self.assertTrue(len(readout["hist_value"]) == len(readout["hist_variance"]))
+        self.assertTrue(len(readout["hist_value"]) == len(readout["dataset"]))
+        self.assertTrue(len(readout["hist_value"]) == len(readout["flavor"]))
+        self.assertTrue(len(readout["hist_value"]) == len(readout["eta"]))
+        self.assertTrue(len(readout["hist_value"]) == len(readout["pt"]))
+        self.assertTrue(len(readout.keys()) == 6)
+
+        ## Histograms with alternate types
+        # Simple double
+        h_double = hist.Hist(hist.axis.Regular(10, 0, 1, name="x"))  # Double
+        readout = read_hist(h_double)
+        self.assertTrue(len(readout["hist_value"]) == len(readout["x"]))
+        self.assertTrue(len(readout.keys()) == 2)
+
+        h_mean = hist.Hist(
+            hist.axis.Regular(10, 0, 1, name="x"), storage=hist.storage.Mean()
+        )
+        readout = read_hist(h_mean)
+        self.assertTrue(len(readout["hist_value"]) == len(readout["x"]))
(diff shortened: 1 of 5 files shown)


clelange:
Thanks, this looks useful! I've triggered the tests, but we should have tests as you also point out to make sure this is working and will continue to work in the future. I'll wait for you to add some. Let me know if you need help.

yimuchen:
Cool! If the developers think this contribution is in the right direction, I will continue rounding out the features, testing and the linting! More to come over the next few days.

yimuchen:
@clelange I have included all the features that for testing. Also checked with against `pylint` for the new `hist_utils.py`.

GraemeWatt:
@yimuchen : can you please also check the linting of `tests/test_histread.py`?  The CI runs `python -m pylint tests/*.py --rcfile=tests/pylintrc` which is currently failing.

tests/test_histread.py
```diff
@@ -47,8 +47,10 @@ def test_default_read(self):
         """
         try:
             readout = read_hist(TestHistUtils.base_hist)
+        # pylint: disable=W0702
```
clelange, line 50: Why are you using this `try`-`except` block here (and also below) in the first place? Does it ever fail? If so, you should know which kind of exception is thrown and use that. Please do not disable pylint warnings.

yimuchen:
Hm... I just checked, the `hist` package is only strictly available for `python>=3.7`, so that is why the final scripts are failing. I'm not sure if there is an easy way to limit sub-packages to require a different python version.



[1] https://pypi.org/project/hist/

tests/test_histread.py
yimuchen: This is following the format that was found in `cfilereader` and other tests [1]. 

Hm... I guess since the format parsing is pure python. I should actually remove the try/except all together, and simply have pytest handle the exception?

https://github.com/HEPData/hepdata_lib/blob/main/tests/test_cfilereader.py#L48

clelange: Ah, OK, I had forgotten that this is used elsewhere. I think also in https://github.com/HEPData/hepdata_lib/blob/main/tests/test_cfilereader.py#L48 it should actually not be used, but let's not worry about that here.
I'm wondering what `test_default_read` is meant to test. Can `read_hist(TestHistUtils.base_hist)` actually fail? I would think that it can only happen if the `hist.Hist()` interface changes. So I would think that you can just read in the file and get rid of the `try`-`except`. However, there might be other ways that reading fails. For example, is it possible that an invalid hist is provided?

clelange:
> Hm... I just checked, the `hist` package is only strictly available for `python>=3.7` [1], so that is why some of the CI scripts are failing. I'm not sure if there is an easy way to limit sub-packages to require a different python version.
> 
> [1] https://pypi.org/project/hist/

Yes, there's a way, but I need to look it up. We could either deprecate Python 3.6 since it's EOL anyway or do that. I'll get back to you later (hopefully tomorrow) in case nobody else has done so in the meantime.

clelange:
> > Hm... I just checked, the `hist` package is only strictly available for `python>=3.7` [1], so that is why some of the CI scripts are failing. I'm not sure if there is an easy way to limit sub-packages to require a different python version.
> > [1] https://pypi.org/project/hist/
> 
> Yes, there's a way, but I need to look it up. We could either deprecate Python 3.6 since it's EOL anyway or do that. I'll get back to you later (hopefully tomorrow) in case nobody else has done so in the meantime.

Thinking about this a bit more, I would say we should just get rid of Python 3.6 support. It's not great since lots of systems we use still have 3.6 as system Python, but then again there are relatively easy ways to get to Python >= 3.7 and we need to make sure we document them.

GraemeWatt:
> Hm... I just checked, the `hist` package is only strictly available for `python>=3.7` [1], so that is why some of the CI scripts are failing. I'm not sure if there is an easy way to limit sub-packages to require a different python version.
> 
> [1] https://pypi.org/project/hist/

The last version of `hist` for Python 3.6 is [v2.4.0](https://pypi.org/project/hist/2.4.0/).  The tests are failing for Python 3.6 not because `hist` cannot be installed, but because you're using features from `hist` that were not available in v2.4.0.  So it might be possible to keep Python 3.6 support if you rewrite your code to work with `hist` v2.4.0.  But if this is difficult or you need features from the later `hist` versions, then dropping Python 3.6 support is definitely the simpler solution.

yimuchen:
Let me quickly compare what are the minimum set of features required for hist==2.4.0

yimuchen:
This new readout method should work with both older versions of hist, as well as user defined accumulators.

clelange (approved):
Sorry about the delay, looks good to me. I'll create a new release soon (might be tomorrow though).

Issue #247, opened 2024-02-07
Allow dummy uncertainties for variable member with dummy (string) value


wmtford:
In constructing HEPData tables, sometimes we have distributions in which no actual information is available for a subset of the dependent variables in some bins.  One solution I’ve used for these is to enter a string, like ‘---’ for the value and zero for the uncertainty.  This works, but with lots of warnings about zero error values.  Would it be possible to protect string values of the uncertainties, as is done for the central values?  Then one could enter a similar dummy indicator string for those.
 
Of course, the consumer of the table has to trap any special dummy contents in the “values” and “uncertainties” fields.  I don’t know if there are any guidelines for conventions to make this uniform.

clelange:
Hi Bill, thanks for creating this issue. Do you have an example of how you solved this that you could share so that it's easier to test and reproduce (and then to change eventually)?

GraemeWatt:
Thanks for the feedback, but HEPData doesn't support string values for the uncertainties.  Recommendations for how to encode missing bins are given in:

  https://hepdata-submission.readthedocs.io/en/latest/data_yaml.html#uncertainties  (second paragraph)

The implementation in `hepdata_lib` was done in PR #161, i.e. if all uncertainties are zero for a particular bin, then the `errors` key is omitted from the YAML output.  The warning message was [my suggestion](https://github.com/HEPData/hepdata_lib/pull/161#discussion_r616603310) to discourage empty bins if there is only one dependent variable.  But for your use case where there are multiple dependent variables and a (different) subset of bins are empty for each dependent variable, it is legitimate to set uncertainties to zero for those bins and you should just ignore the warnings from `hepdata_lib`.

I don't think the behaviour of the `hepdata_lib` code needs to be changed, but there could be better [documentation](https://hepdata-lib.readthedocs.io/en/latest/usage.html#uncertainties) and perhaps an option could be added to suppress the warnings.  I'll try to address these two points and I'll leave this issue open until they are addressed.

wmtford:
Thanks both.  An example of the use case is Fig. 5, the second table, in [1].  A stand-alone notebook to reproduce that table, along with the needed input files, can be found in cernbox at [2].  From my perspective this works fine, other than that the warning messages seem to imply a problem.  

The reason that the encoding of missing bins pointed to by Graeme doesn't work in the hepdata_lib implementation is that there uncertaintes are added, or not, for an entire column; we don't have access to treat some of the rows differently.
 
[...]

PR #251, opened 2024-02-13
Clarify behaviour if zero uncertainties

GraemeWatt:
* Add a [paragraph](https://hepdata-lib--251.org.readthedocs.build/en/251/usage.html#uncertainties) to the documentation to explain how to encode missing bins.
* Add an option `zero_uncertainties_warning` to the `Variable` class with default value `True`.
* Add a test for a data point with zero uncertainties.  This should have been added as part of PR #161.
* In response to #247.

----
📚 Documentation preview 📚: https://hepdata-lib--251.org.readthedocs.build/en/251/

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -337,4 +337,12 @@ After creating the Uncertainty objects, the only additional step is to attach th
     variable.add_uncertainty(unc1)
     variable.add_uncertainty(unc2)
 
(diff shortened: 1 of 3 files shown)


clelange (approved):
Thanks, looks good to me! I'm just wondering whether at some point `hepdata_lib` should actually support this, so we can keep the issue open for now.

PR #254, opened 2024-02-14
Update and move SWAN-specific instructions

GraemeWatt:
The "SWAN-specific instructions" at the top of [Getting_started.ipynb](https://github.com/HEPData/hepdata_lib/blob/main/examples/Getting_started.ipynb) are misleading as they imply that `hepdata_lib` needs to be installed.  Moreover, the given command `pip install --user hepdata_lib` does not work to upgrade `hepdata_lib` without setting a `$PYTHONPATH` variable in a startup script.  See an email I sent in response to a user query on 22nd January 2024.  I updated the "SWAN-specific instructions" and moved them from the `Getting_started.ipynb` notebook to a new "[Using SWAN](https://hepdata-lib--254.org.readthedocs.build/en/254/setup.html#using-swan)" section in `docs/setup.rst`.  I also printed the `hepdata_lib` version at the top of the `Getting_started.ipynb` notebook so that users can more easily see which version is installed.

If we make sure that the LCG Releases are updated regularly with new `hepdata_lib` releases, it should not be necessary for users to upgrade `hepdata_lib` on their own CERNBox, given that the procedure is not completely straightforward. @clelange : when you make a new `hepdata_lib` release, could you please also open a JIRA ticket (https://its.cern.ch/jira/projects/SPI/) to request that the new version be included in the next LCG release?  I can do this if you prefer, but it would make sense to do it as part of the `hepdata_lib` release procedure.

----
📚 Documentation preview 📚: https://hepdata-lib--254.org.readthedocs.build/en/254/

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ If you are not sure about your Python environment, please also see below how to 
 
 ## Getting started
 
-For using `hepdata_lib`, you don't even need to install it, but can use the [binder](https://mybinder.org/) or [SWAN](https://swan.cern.ch/) (CERN-only) services using one of the buttons below and following the instructions in the notebook with name [Getting_started](examples/Getting_started.ipynb):
+For using `hepdata_lib`, you don't even need to install it, but can use the [binder](https://mybinder.org/) or [SWAN](https://swan.cern.ch/) (CERN-only) services using one of the buttons below:
 
 [![Binder](https://mybinder.org/badge.svg)](https://mybinder.org/v2/gh/HEPData/hepdata_lib/main?filepath=examples/Getting_started.ipynb)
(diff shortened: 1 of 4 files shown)


clelange:
Thanks for the updated documentation @GraemeWatt -- would you mind adding to the development docs that the JIRA issue should be created (after the PyPI upload)? I've created a JIRA issue just now for the release from a few days ago: https://its.cern.ch/jira/browse/SPI-2507

GraemeWatt:
> Thanks for the updated documentation @GraemeWatt -- would you mind adding to the development docs that the JIRA issue should be created (after the PyPI upload)? I've created a JIRA issue just now for the release from a few days ago: https://its.cern.ch/jira/browse/SPI-2507

Now added:  https://hepdata-lib--254.org.readthedocs.build/en/254/dev.html#making-a-release

clelange (approved):
Thank you very much!

Issue #245, opened 2023-12-11
Add ability to specify license information


GraemeWatt:
The JSON schema defining a HEPData submission has provision for a [`data_license`](https://github.com/HEPData/hepdata-validator/blob/bfcd4fd2c3bec027816944a60ca92a270e1c356b/hepdata_validator/schemas/1.1.1/submission_schema.json#L31-L37) (comprising a `name`, `url` and optional `description`) for individual tables.  It also supports a [`license`](https://github.com/HEPData/hepdata-validator/blob/bfcd4fd2c3bec027816944a60ca92a270e1c356b/hepdata_validator/schemas/1.1.1/additional_resources_schema.json#L12-L18) for additional resources.  An example of use is given in [`submission.yaml`](https://github.com/HEPData/hepdata-submission/blob/main/examples/submission/TestHEPSubmission/submission.yaml) used in [`TestHEPSubmission.zip`](https://github.com/HEPData/hepdata-submission/blob/main/examples/submission/TestHEPSubmission.zip) and in the [submission documentation](https://hepdata-submission.readthedocs.io/en/latest/submission_yaml.html#full-example).  By default, the [CC0](https://creativecommons.org/publicdomain/zero/1.0/legalcode) license applies, as explained in the [HEPData Terms of Use](https://www.hepdata.net/terms).

The [`add_additional_resource`](https://github.com/HEPData/hepdata_lib/blob/f194e14928538a3f62a6e6373e0df03d59bee23c/hepdata_lib/__init__.py#L58C9-L58C32) function of `hepdata_lib` should be extended to accept an optional `license` argument (a Python dictionary containing `name`, `url` and optional `description`).  The `Table` class should also be extended to include a new method `add_data_license` taking arguments `name`, `url` and optional `description`.

See also HEPData/hepdata#740.

GraemeWatt:
Documentation should be added to `hepdata_lib` explaining how to add a license to either an additional resource file or a data table.  As an example, the [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/) license could be used.
```yaml
name: CC BY 4.0
url: https://creativecommons.org/licenses/by/4.0/
description: This license enables reusers to distribute, remix, adapt, and build upon the material in any medium or format, so long as attribution is given to the creator.
```

Issue #234, opened 2023-10-14
Upgrade Pylint to allow use with Python 3.11


GraemeWatt:
Currently, Pylint is pinned at v2.9.6, which installs [wrapt](https://github.com/GrahamDumpleton/wrapt) v1.12.1.  Trying to use Pylint v2.9.6 with Python 3.11 gives an error message [`ImportError: cannot import name 'formatargspec' from 'inspect'`](https://github.com/GrahamDumpleton/wrapt/issues/196).  Upgrading to Pylint v2.15 or later fixes the problem.  However, running a later version of Pylint gives messages like `unrecognized-option` and `useless-option-value`.  The `pylintrc` and `tests/pylintrc` files need to be revised, then Pylint can be upgraded and Python 3.11 can be added to the test matrix in the CI (`tests.yml`).

PR #257, opened 2024-02-27
Data licence addition

ItIsJordan:
* Adds functionality to insert resource file licence information into HEPData submissions in the `add_additional_resource()` function.
* Also adds a new `add_data_license()` function to the `Table` class.
* Updates documentation and testing for new features.
* Changes default license written by the `get_license()` function from [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/) to [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
* Closes #257.

----
📚 Documentation preview 📚: https://hepdata-lib--257.org.readthedocs.build/en/257/


ItIsJordan:
This PR is intended to close #245.

ItIsJordan:
I just noticed the coverage. Sorry, will look into that this week.

GraemeWatt:
@clelange : when testing Jordan's branch, I noticed that `hepdata_lib` already defines a default license:

https://github.com/HEPData/hepdata_lib/blob/ab8bbef609eaa8b012612520646825530ca50e97/hepdata_lib/__init__.py#L509-L517

and writes it as `data_license` in the first YAML document of the `submission.yaml` file:

https://github.com/HEPData/hepdata_lib/blob/ab8bbef609eaa8b012612520646825530ca50e97/hepdata_lib/__init__.py#L613

This has been the behaviour ever since the first `hepdata_lib` version from 2018.  However, `data_license` has never been supported in the relevant [`additional_info_schema.json`](https://github.com/HEPData/hepdata-validator/blob/main/hepdata_validator/schemas/1.1.1/additional_info_schema.json).  The JSON schema only support a `data_license` in the [`submission_schema.json`](https://github.com/HEPData/hepdata-validator/blob/bfcd4fd2c3bec027816944a60ca92a270e1c356b/hepdata_validator/schemas/1.1.1/submission_schema.json#L31-L40) (data tables) or a `license` in the [`additional_resources_schema.json`](https://github.com/HEPData/hepdata-validator/blob/bfcd4fd2c3bec027816944a60ca92a270e1c356b/hepdata_validator/schemas/1.1.1/additional_resources_schema.json#L12-L18) (additional resources).  It would be difficult to modify our database model to support a license for a whole submission rather than just individual tables and resource files.  Moreover, the default `hepdata_lib` license (CC BY 4.0) is slightly more restrictive than the default HEPData license (CC0).

Therefore, I removed the writing of a default license in commit 2cf90464ec5a05e122504019d37b444da6890e51.  We could revert that commit if you disagree with the removal.  However, I think it will be confusing if we provide new methods to write license information for data tables and additional resources, but keep a redundant default license for the whole submission that will just be ignored when uploading to HEPData.

clelange:
@GraemeWatt - I think there should be a default license that is applied, and I agree that this should be CC0 since that's the default stated on the HEPData web page. I don't remember why we chose CC BY 4.0 back then. As part of this PR, I think it would be good to state in the documentation that this license is applied by default and that this is the default for all HEPData records as well as mentioning that users can change to their own license (with the newly added functionality of this PR).

GraemeWatt:
Looks like the CI has started failing when building a wheel for `wrapt` v1.12.1, installed by `astroid` v2.6.6 via `pylint` v2.9.6.  One solution might be to address #234 by upgrading `pylint` to a more recent version.

GraemeWatt:
I opened a PR #260 that removes the pin on Pylint and fixes the build, so better to wait for that PR to be merged.

clelange:
Looks like this is now good to go? @ItIsJordan

ItIsJordan:
Should be ready now, yes! @clelange

examples/Getting_started.ipynb
```diff
@@ -265,25 +265,50 @@
     "If you want, the original data file can be attached to the table as an additional resource file."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The default license for HEPData records is [CC0](https://creativecommons.org/publicdomain/zero/1.0/legalcode) (see [Terms of Use](https://www.hepdata.net/terms)), but it is possible to specify a different license for both additional resource files and data tables."
```
clelange, line 272: I would actually suggest not to have this in the "Getting started" notebook since this is what people will typically copy and paste, and in this case they would use a more restrictive license...
Could you please remove those lines again?

hepdata_lib/__init__.py
```diff
@@ -505,17 +565,6 @@ def __init__(self):
             "Created with hepdata_lib " + __version__,
             "https://doi.org/10.5281/zenodo.1217998")
 
-    @staticmethod
-    def get_license():
```
clelange, line 509: By removing this, by default there won't be a license attached to the record anymore? I think this isn't good practice, the records themselves should always come with a license, even if somewhere on the portal it is stated that there's a default. My suggestion would be to add by default always the default license (which would need to be updated to CC0).
@GraemeWatt What do you think?

clelange (commented):
I left two comments for discussion that I think are important.

examples/Getting_started.ipynb
GraemeWatt: OK, I agree that addition of license information is an advanced feature that should not be included in the "Getting started" notebook, so I removed it.  I think part of my original intention was to find a simple way to increase the code coverage before Jordan expanded the tests.

clelange (approved):

hepdata_lib/__init__.py
GraemeWatt: OK, I reverted commit 2cf90464ec5a05e122504019d37b444da6890e51 and changed the default license to CC0 in commit 2ce00358975563b445005a924106bbbe73b55374.  However, as explained in my [original comment](https://github.com/HEPData/hepdata_lib/pull/257#issuecomment-1967337307), a `data_license` written as the first document of the `submission.yaml` file is not supported by the JSON schema and it will not be processed by the HEPData submission code.  Licenses are only supported for individual data tables and resource files, not for whole submissions.  The default license will be visible if a user downloads the YAML format, however.

clelange: Thanks, I hope I didn't merge too soon.

PR #258, opened 2024-02-28
Update usage.rst to include self-referencing notice

ItIsJordan:
Updates the bidirectional linking documentation for records/tables to note that the intended use for the field is not for self-referencing as mentioned in [HEPData/hepdata Issue 765](https://github.com/HEPData/hepdata/issues/765)

----
📚 Documentation preview 📚: https://hepdata-lib--258.org.readthedocs.build/en/258/


clelange (approved):
Thanks!

PR #259, opened 2024-03-15
Add `try`/`except` around ROOT import

### LINKED ISSUE #99: Add flag to disable features requiring ROOT
There are use cases where the tool is used but no ROOT installation is available. We should add an option that disables all features that use ROOT to make the other features work independently (e.g. reading in a text file). For ROOT TTrees, we could fall back to uproot.

GraemeWatt:
* Print out an error message if ROOT cannot be imported.
* Add some explanation at the end of the README.md file.
* Correct a typo in the link to `reading_scikithep_histograms.ipynb`.
* Add a Pytest [custom marker](https://docs.pytest.org/en/7.1.x/example/markers.html) to test functions requiring ROOT.
* Allow CI to run subset of tests without ROOT using `pytest -m 'not needs_root'`.
* Closes #99.

----
📚 Documentation preview 📚: https://hepdata-lib--259.org.readthedocs.build/en/259/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        root-version: ["6.24", "6.26", "6.28", "6.30"]
+        root-version: ["", "6.24", "6.26", "6.28", "6.30"]
         python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
         exclude:
           - root-version: "6.24"
@@ -55,6 +55,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Setup Micromamba environment
+      if: ${{ matrix.root-version }}
       uses: mamba-org/setup-micromamba@v1
       with:
         environment-name: ci
@@ -68,7 +69,22 @@ jobs:
           channels:
             - conda-forge
 
+    - name: Setup Micromamba environment without ROOT
+      if: ${{ !matrix.root-version }}
+      uses: mamba-org/setup-micromamba@v1
+      with:
+        environment-name: ci
+        create-args: >-
+          python=${{ matrix.python-version }}
+          imagemagick
+          ghostscript
+          pip
+        condarc: |
+          channels:
+            - conda-forge
+
     - name: ROOT info
+      if: ${{ matrix.root-version }}
       run: |
         root-config --version
         root-config --python-version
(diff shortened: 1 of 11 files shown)


GraemeWatt:
@clelange @alexander-held : I made a crude but simple attempt in this PR to address #99 by only importing ROOT functions in the `hepdata_lib` functions where they are actually needed.  This means that `import hepdata_lib` now works without a ROOT installation as does the basic functionality (e.g. demonstrated in the [Getting_started.ipynb](https://github.com/HEPData/hepdata_lib/blob/main/examples/Getting_started.ipynb) notebook).

alexander-held:
This is very nice to see, thanks for taking care of this!

clelange:
Hi @GraemeWatt -- thanks for this! I'm wondering, however, if for `hepdata_lib/root_utils.py` (and probably also `hepdata_lib/c_file_reader.py`) it's not OK to actually keep the top-level ROOT import in a `try`-`catch` statement and rather print out a warning if the ROOT import failed.

My main issue with these changes is that while they move some of the ROOT import failures to later and therefore `import hepdata_lib` will not fail immediately, we currently don't have visibility how this would fail. It would be nice to actually have tests without a ROOT installation and then label in the docs which features work without a ROOT installation.

What do you think?

GraemeWatt:
@clelange : Yes, I agree that keeping the top-level ROOT import and adding a `try`/`except` condition is a better (and simpler) solution than my first attempt at only importing ROOT functions in the `hepdata_lib` functions where they are actually needed.  I've added a new commit on top of the `upgrade-pylint` branch and force-pushed to the `change-root-import` branch, so you should merge PR #260 before this one.  I've added an explanation at the end of the `README.md` file.  I don't see an easy way to modify the CI to work without a ROOT installation, but I tested locally (in a Python 3.11 virtual environment with an incompatible ROOT built with Python 3.12) that the code behaves as expected.  Feel free to expand the tests run by the CI if you think it is important and you can work out how to do it.

clelange:
> Feel free to expand the tests run by the CI if you think it is important and you can work out how to do it.

I think we can just run some of the tests without installing ROOT, i.e. a separate set of tests. However, we can do that in a later PR.

There's a conflict for `hepdata_lib/c_file_reader.py`. Could you please check if you can fix it? You could also rebase on top of the current `main` branch now that #260 is merged, whichever is easier for you.

GraemeWatt:
@clelange : I used a Pytest [custom marker](https://docs.pytest.org/en/7.1.x/example/markers.html) to add a marker `needs_root` to the tests that require a ROOT installation.  I then introduced an additional element to the CI matrix without a ROOT installation `root-version = ""` and ran the tests using `pytest -m 'not needs_root'`.  I also resolved the conflict with the `main` branch.

clelange (approved):
Awesome, thanks, this is a very nice solution!

PR #260, opened 2024-03-28
Remove pin on Pylint to support Python 3.11

### LINKED ISSUE #234: Upgrade Pylint to allow use with Python 3.11

GraemeWatt:
* Remove "pylintrc" and "tests/pylintrc" configuration files.
* Tweaked code so that Pylint checks all pass.
* Add Python 3.11 and ROOT 6.30 to matrix in tests.yml.
* Skip Pylint step in CI for Python 3.6 and Python 3.7, since these Python versions are incompatible with Pylint 3.x.
* ROOT binary built with Python 3.12 not yet supported by Micromamba, so don't include Python 3.12 in test matrix.
* Closes #234 and fixes the build currently broken in PR #257.

----
📚 Documentation preview 📚: https://hepdata-lib--260.org.readthedocs.build/en/260/


tests/pylintrc
clelange, line 1: Could we actually try to remove this file and use only a single `pylintrc`?

clelange (commented):
Thanks, it's very useful to remove the pin on pylint! I left two comments for your consideration before merging this.

pylintrc
```diff
@@ -1,290 +1,324 @@
...
 load-plugins=
 
 # Pickle collected data for later comparisons.
 persistent=yes
 
-# Specify a configuration file.
-#rcfile=
+# Minimum Python version to use for version dependent checks. Will default to
+# the version used to run pylint.
+py-version=3.12
```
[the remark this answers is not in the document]
GraemeWatt: I guess it should be `py-version=3.8` since that's the minimum Python version used in the CI to run Pylint.  Running `pylint --generate-rcfile` in a Python 3.10 or 3.11 virtual environment on my laptop gives `py-version=3.12` for an unknown reason contrary to the [docs](https://pylint.readthedocs.io/en/latest/user_guide/configuration/all-options.html#py-version) which say that the default value is `sys.version_info[:2]`.  Maybe `pylint` somehow uses my default system installation of Python 3.12 (?).  In any case, I removed `pylintrc` in my last commit.

tests/pylintrc
GraemeWatt: On reflection, the `pylintrc` files mostly contain default options, so I decided just to remove them completely and address the remaining issues flagged by running `pylint`:

1. Address "consider-using-f-string" by using Python f-strings wherever `pylint` complained.
2. Address "unspecified-encoding" by adding `encoding='utf-8'` to `open` commands wherever `pylint` complained.
3. Added a comment `# pylint: disable=no-name-in-module` to relevant lines for ROOT imports.
4. Added a comment `# pylint: disable=no-member` to relevant lines using ROOT objects.
5. Expanded tests to avoid Codecov complaints about the modified lines being uncovered.

I also modified the `README.md` file to mention that Python 3.11 and Python 3.12 is supported.  We can't use Python 3.12 in the CI because a [Conda](https://anaconda.org/conda-forge/root/) package is not yet available for ROOT with Python 3.12, but I've checked on my laptop that the tests pass with Python 3.12 (using ROOT with Python 3.12 installed via Homebrew).

clelange: Excellent, thanks a lot!

clelange (approved):

Issue #229, opened 2023-07-06
Add ability to read in YODA files


GraemeWatt:
For cases where an analyser has data already in the [YODA](https://yoda.hepforge.org) format for use with [Rivet](https://yoda.hepforge.org), it would be useful if `hepdata_lib` could read YODA files for conversion to the HEPData YAML format.  It would be preferable if YODA was an optional and not mandatory dependence.  The question of converting YODA to HEPData YAML has been a long-standing issue (HEPData/hepdata-converter#10), but it would be better handled by `hepdata_lib` than the `hepdata-converter`.

Cc: @20DM

20DM:
Hi Graeme!

I'm just in the process of preparing submissions for the reference data files in Rivet that don't have a HepData entry yet. I'm currently struggling to use the `hepdata_lib` for cases with inhomogeneous error breakdowns across bins. For instance, I have a distribution with three bins where the first two bins have two error components 'A' and 'B' (but not 'C') and the third bin has error component 'C' (but not 'A' and 'B').

I know this is supported in principle, e.g. by just omitting the respective components in the dictionary. However, when using the library, it seems `hepdata_lib/helpers.py` raises a `ValueError` 

```
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.
```
Is there a trick?

20DM:
PS - just to be clear: of course I can "make it pass" by just setting the uncertainty to zero, but then all bins will have three uncertainty components, some of them zero, which is not the same as the bin not having the component in its breakdown to begin with. I think the problem is that the check for "non-zero uncertainties" only checks if there's at least one non-zero component and then adds all of them, regardless of their value. Can we make this more flexible?

20DM:
On second thought, I suspect this requirement comes from the cases where we have a differential distribution, which is prepended/appended by a single bin corresponding to the average, which probably shouldn't be allowed. Maybe best to leave the validator as is and I will work around these cases (there's only 5 of them, so should be manageable).

GraemeWatt:
[...]

PR #265, opened 2024-07-22
Add support for inhomogenous error breakdowns

### LINKED ISSUE #266: Support inhomogeneous uncertainties
> I'm currently struggling to use the `hepdata_lib` for cases with inhomogeneous error breakdowns across bins. For instance, I have a distribution with three bins where the first two bins have two error components 'A' and 'B' (but not 'C') and the third bin has error component 'C' (but not 'A' and 'B').
>
> I know this is supported in principle, e.g. by just omitting the respective components in the dictionary. However, when using the library, it seems `hepdata_lib/helpers.py` raises a `ValueError` 
>
> ```
> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.
> ```
> Is there a trick?

_Originally posted by @20DM in https://github.com/HEPData/hepdata_lib/issues/229#issuecomment-2242550182_

> PS - just to be clear: of course I can "make it pass" by just setting the uncertainty to zero, but then all bins will have three uncertainty components, some of them zero, which is not the same as the bin not having the component in its breakdown to begin with. I think the problem is that the check for "non-zero uncertainties" only checks if there's at least one non-zero component and then adds all of them, regardless of their value. Can we make this more flexible?

_Originally posted by @20DM in https://github.com/HEPData/hepdata_lib/issues/229#issuecomment-2242610663_

20DM:
Proposal solution for the feature discussed in #229 :

The HepData validator already allows for cases where different bins have different error breakdowns, but `hepdata_lib` currently does not support this.

With this change set one can suppress individual sources by setting them to `None`.

Closes #266.

----
📚 Documentation preview 📚: https://hepdata-lib--265.org.readthedocs.build/en/265/

### DIFF
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -274,7 +274,7 @@ def sanitize_value(value):
     Handle conversion of input types for internal storage.
 
     :param value: User-side input value to sanitize.
-    :type value: string, int, or castable to float
+    :type value: string, int, NoneType, or castable to float
 
     Strings, integers and None are left alone,
     everything else is converted to float.
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -283,6 +283,8 @@ def sanitize_value(value):
         return value
     if isinstance(value,int):
         return value
+    if value is None:
+        return value
     return float(value)
 
 
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -276,7 +276,7 @@ def sanitize_value(value):
     :param value: User-side input value to sanitize.
     :type value: string, int, or castable to float
 
-    Strings and integers are left alone,
+    Strings, integers and None are left alone,
     everything else is converted to float.
     """
     if isinstance(value,str):
--- a/hepdata_lib/helpers.py
+++ b/hepdata_lib/helpers.py
@@ -258,7 +258,8 @@ def any_uncertainties_nonzero(uncertainties, size):
     for unc in uncertainties:
 
         # Treat one-sided uncertainties as
-        values = np.array(unc.values)
+        tmp = 0 if unc.is_symmetric else (0,0)
+        values = np.array([tmp if v is None else v for v in unc.values])
         values[values.astype(str)==''] = 0
         values = values.astype(float)
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -370,5 +370,7 @@ multiple dependent variables and a (different) subset of the bins has missing co
 In this case the uncertainties should be set to zero for the missing bins with a non-numeric central value like ``'-'``.
 The warning message can be suppressed by passing an optional argument ``zero_uncertainties_warning=False`` when
 defining an instance of the ``Variable`` class.
+Furthermore, note that `None` can be used to suppress the uncertainty for individual bins in cases where
+the uncertainty components may only apply to a subset of the values.
 
 .. _`Uncertainties`: https://hepdata-submission.readthedocs.io/en/latest/data_yaml.html#uncertainties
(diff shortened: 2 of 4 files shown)


GraemeWatt (changes requested):
Thanks for the PR!  Can you please add a test in `tests/test_uncertainty.py` to cover the new feature and add some explanation under [Uncertainties](https://github.com/HEPData/hepdata_lib/blob/main/docs/usage.rst#uncertainties)? in the documentation?

GraemeWatt:
Thanks for adding the test and expanding the docs.  The CI is failing due to issues with `pylint`.  Can you please modify the code so that the `pylint` checks pass?  See [Analysing the code](https://hepdata-lib.readthedocs.io/en/latest/dev.html#analysing-the-code) in the developer docs.

hepdata_lib/helpers.py
```diff
@@ -282,6 +283,8 @@ def sanitize_value(value):
         return value
     if isinstance(value,int):
         return value
+    if value is None:
+        return value
```
20DM, line 287: Ouff, I will take a look!

I just wanted to highlight this part of the change set, which I had to put in to make the unit test pass, but it may not be what we want in general (i.e. for non-Uncertainty-type classes)? I didn't see this in my local tests at first, because the sanitising function isn't called when appending to the values array as part of a loop.

GraemeWatt: I think that's fine and it shouldn't have unintended consequences, since previously `sanitize_value(None)` would have just given an exception from `float(None)`.  The docstring of `sanitize_value` should be modified, however, to mention that a `value` of `None` is now allowed.

GraemeWatt (changes requested):
Thanks for fixing the linting for the tests.  However, the CI still gives a message:
```
************* Module hepdata_lib
hepdata_lib/__init__.py:237:4: R0912: Too many branches (13/12) (too-many-branches)

-----------------------------------
Your code has been rated at 9.99/10
```
Could you please fix this too?  If you don't see an easy way to refactor the `make_dict` function to avoid this warning, it can be suppressed by adding a line `# pylint: disable=too-many-branches` at the top of the function, but this should be avoided if possible.  I think it's a current bug in the CI that the "Run pylint" still passes, so I've opened a separate PR #267 to address it.

20DM:
> Thanks for fixing the linting for the tests. However, the CI still gives a message:
> 
> ```
> ************* Module hepdata_lib
> hepdata_lib/__init__.py:237:4: R0912: Too many branches (13/12) (too-many-branches)
> 
> -----------------------------------
> Your code has been rated at 9.99/10
> ```
> 
> Could you please fix this too? If you don't see an easy way to refactor the `make_dict` function to avoid this warning, it can be suppressed by adding a line `# pylint: disable=too-many-branches` at the top of the function, but this should be avoided if possible. I think it's a current bug in the CI that the "Run pylint" still passes, so I've opened a separate PR #267 to address it.

Urgh - well, this can be circumvented by moving the additional if statement into a separate list comprehensions, e.g. like so
```
for unc in [ unc for unc in self.uncertainties if not unc.values[i] is None ]:
    ...
```
but that adds an extra loop for no good reason. So my preference would be to silence the linter here, but I'd leave that call to you. 🤷

hepdata_lib/helpers.py
20DM: Thanks - done now!

GraemeWatt (approved):
I've added `# pylint: disable=too-many-branches` at the top of the `make_dict` function, added `NoneType` in the docstring of the `sanitize_value` function, and merged the last commit to the `main` branch.  The [workflow run](https://github.com/HEPData/hepdata_lib/actions/runs/10060011699) for the separate "Run pylint on hepdata_lib" step now fails before adding the `# pylint: disable=too-many-branches` line, as expected.  I'll let @clelange perform the final review and merge.

clelange (approved):
Thanks for the contribution @20DM and for the review @GraemeWatt !

PR #273, opened 2024-10-04
Require named arguments to fix pylint too-many-positional-arguments

### LINKED ISSUE #272: Tests fail with pylint >= 3.3.0
Errors:

```
hepdata_lib/__init__.py:58:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments)
hepdata_lib/__init__.py:146:4: R0917: Too many positional arguments (7/5) (too-many-positional-arguments)
```

This seems to be due to the new feature introduced in https://github.com/pylint-dev/pylint/releases/tag/v3.3.0:

> Added `too-many-positional-arguments` to allow distinguishing the configuration for too many total arguments (with keyword-only params specified after `*`) from the configuration for too many positional-or-keyword or positional-only arguments.

> As part of evaluating whether this check makes sense for your project, ensure you adjust the value of `--max-positional-arguments`.

clelange:
Fixes #272 

----
📚 Documentation preview 📚: https://hepdata-lib--273.org.readthedocs.build/en/273/

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -55,7 +55,7 @@ class AdditionalResourceMixin:
         self.files_to_copy = []
         self.additional_resources = []
 
-    def add_additional_resource(self, description, location, copy_file=False, file_type=None,
+    def add_additional_resource(self, description, location, *, copy_file=False, file_type=None,
                                 resource_license=None):
         """
         Add any kind of additional resource.
@@ -143,7 +143,7 @@ class Variable:
     # pylint: disable=too-many-instance-attributes
     # Eight is reasonable in this case.
 
-    def __init__(self, name, is_independent=True, is_binned=True, units="", values=None,
+    def __init__(self, name, *, is_independent=True, is_binned=True, units="", values=None,
                  zero_uncertainties_warning=True):
         # pylint: disable=too-many-arguments
         self.name = name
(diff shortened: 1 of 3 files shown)


clelange:
This needs some more fixes in other places, I had only addressed the errors that showed up first in the tests.

GraemeWatt:
@clelange : thanks for addressing #272.  I noticed the problem yesterday, but I didn't get around to opening a PR.  Although it's better programming practice, it's a breaking change to reassign some arguments from positional to keyword-only.  Do you anticipate this will break user code?   It might be better in the long run to reassign some arguments and I leave the decision to you, but a simple fix would be to add `# pylint: disable=too-many-positional-arguments` to the offending functions.

clelange:
Hi @GraemeWatt -- while making some of the changes, I was also thinking about whether this breaks user code or not. So far, it looks like it's only smaller things that would be easy to fix. Forcing more named function parameters would actually make the code better readable and usable, so I'm somewhat in favour of implementing the fixes rather than silencing the linter.

clelange:
It really looks like the changes are minor and good to have even though it might break for some edge cases where people don't like named arguments.

GraemeWatt (approved):
Thanks, this looks good.  I agree that it is better practice not to allow keyword arguments to be given as positional arguments.

Issue #276, opened 2025-05-09
Import ROOT functions only where needed?


GraemeWatt:
[Feedback](https://gitlab.com/hepcedar/yoda/-/merge_requests/329#note_2494375578) from @20DM:

> I'm working on adding support for converting YODA files into HepData draft submissions, and I noticed that importing the `hepdata_lib` module prints the following message:
> ```
> Cannot import ROOT: No module named 'ROOT'
> ```
> This message appears every time `import hepdata_lib` is called in a script, which feels a bit too verbose - especially since ROOT isn't always needed.
>
> Would it be possible to defer this warning until one of the ROOT-dependent methods is actually used? That way, users who don’t need ROOT won't see unnecessary warnings just from importing `hepdata_lib`. Thanks!

That was actually my original way of addressing issue #99 via PR #259 (original title: "Import ROOT functions only where needed"), but I changed to the top-level ROOT import with a `try`/`except` condition after feedback from @clelange (see discussion in PR #259).  The original code is difficult to find since I force-pushed my branch, but see [here](https://github.com/HEPData/hepdata_lib/compare/5228276cb4ec14ca12559e7643031ae2d09af177..2e77ee8566fdf251be0bacc75156720e9f12e987).

@clelange, given that I later addressed your concern by adding tests without a ROOT installation, do you think we can import ROOT functions only where needed, or do you still prefer to keep the top-level ROOT import?

clelange:
Hi @GraemeWatt --- fine with me to only import where needed.

PR #277, opened 2025-05-14
Remove warning importing ROOT in `root_utils.py` and expand test matrix

GraemeWatt:
* Closes #276.
* The test matrix is also expanded to include ROOT 6.32/6.34 and Python 3.12/3.13.

----
📚 Documentation preview 📚: https://hepdata-lib--277.org.readthedocs.build/en/277/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -76,8 +76,8 @@ jobs:
             python-version: "3.8"
         include:
           - os: macos-latest
-            root-version: "6.30"
-            python-version: "3.11"
+            root-version: "6.34"
+            python-version: "3.13"
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -170,11 +170,11 @@ jobs:
         path: examples/*.html
 
     - name: Run pylint on hepdata_lib
-      if: ${{ always() && matrix.root-version && !startsWith(matrix.python-version, '3.6') && !startsWith(matrix.python-version, '3.7') }}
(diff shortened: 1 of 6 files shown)


GraemeWatt:
@clelange : I'm not sure what (if anything) to do about the [Codecov report](https://app.codecov.io/gh/HEPData/hepdata_lib/pull/277).  I added extra ROOT and Python versions to the test matrix, so the [Unexpected Coverage Changes](https://docs.codecov.com/docs/unexpected-coverage-changes) might just be due to a different number of reports between head (26 uploads) and base (16 uploads).

clelange:
Thanks for this @GraemeWatt !

I'm wondering what we actually gain by performing the imports within functions instead of simply removing the "missing ROOT" warning when importing `ROOT`. Any script trying to use a function that requires ROOT will fail anyway and the error will be similar. Apologies that I didn't think about this when commenting on #276.

The reason why this warning is printed all the time is probably simply the import of `RootFileReader` in https://github.com/HEPData/hepdata_lib/blob/main/hepdata_lib/__init__.py#L23 --- I'm not sure if we actually need it here. So even easier than what I wrote above might be to remove that line and keep the warnings, which are in [`root_utils.py`](https://github.com/HEPData/hepdata_lib/blob/main/hepdata_lib/root_utils.py) and [`c_file_reader.py`](https://github.com/HEPData/hepdata_lib/blob/main/hepdata_lib/c_file_reader.py), which both really need ROOT and therefore the warning is warranted there. Or am I missing something here?

clelange (approved):
Great, thanks a lot!

PR #283, opened 2025-09-25
Remove requirement of the future package

andresailer:
`raise ... from` is available since forever?, so no need for keeping the obsolete future module

The future package relies on 2to3, which was removed in python 3.13. We want to get rid of the future package in the LCG stacks for python 3.13

----
📚 Documentation preview 📚: https://hepdata-lib--283.org.readthedocs.build/en/283/

### DIFF
--- a/hepdata_lib/__init__.py
+++ b/hepdata_lib/__init__.py
@@ -9,7 +9,6 @@ from decimal import Decimal
 from re import match as rematch
 import numpy as np
 import yaml
-from future.utils import raise_from
 
 # try to use LibYAML bindings if possible
 try:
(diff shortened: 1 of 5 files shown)


clelange:
I think we can ignore the coverage output in this case.

andresailer:
Thanks!

Sure I can update hepdata_lib, I added a subscription to the releases so we get notified as well. Not much more than opening a ticket to me is what you can do anyway :)

If you give me an estimate when the next tag will be with this PR included?

clelange:
Then we can also remove `future` from https://github.com/HEPData/hepdata_lib/blob/main/requirements.txt#L3C1-L3C7 ?

clelange:
I created https://github.com/HEPData/hepdata_lib/releases/tag/v0.20.0

clelange:
While more automation would be appreciated, I think the instructions are still OK/correct to make sure this gets into the LCG releases.

andresailer:
> While more automation would be appreciated, I think the instructions are still OK/correct to make sure this gets into the LCG releases.

Yes, we might miss an email, forget, or make a mistake. So if you notice that some update has not been applied, or you just want to make sure the next update happens, please open a JIRA ticket.

